PR #7, opened 2019-01-15
Add support for exposing UITabBarItems to presentTabs

Iamsamwhoami:


Iamsamwhoami:
@isaacSennerholt Updated a bit, code still looks ish the same but there are methods to present somewhere between 1 to 5 tabs.

PR #11, opened 2019-01-17
Logout feature

Iamsamwhoami:
This PR adds a logout button to the profile screen


Hedvig/Assets/Localization/Localization.swift
```diff
@@ -1180,6 +1188,42 @@ struct Translations {
...
                 """
+
+                case .LOGOUT_BUTTON:
+                    if let text = TranslationsRepo.find(.LOGOUT_BUTTON) {
+                        return text
+                    }
+
+                    return """
+                Logga ut
+                """
```
isaacSennerholt: Why so many quotation marks?

Iamsamwhoami: This one is auto-generated, this is the swift way of handling multiline text hehe

PR #23, opened 2019-03-11
Fix payment related bugs

Iamsamwhoami:


Src/Screens/ProfileScreen/Screens/MyPayment/MyPayment.swift
```diff
@@ -76,7 +76,7 @@ extension MyPayment: Presentable {
                 viewController.present(directDebitSetup, options: [.autoPop])
             }
 
-            if result.data?.isSwitchingBankAccount() ?? false {
+            if result.data?.directDebitStatus == .pending {
```
meletios: i think here should be needs_setup

Iamsamwhoami: @meletios this is for hiding the "Connect autogiro"-button, which is only to be hidden if it's pending, right?

meletios: We need to hide it, when its pending or connected, eller?

Iamsamwhoami: only pending :D

meletios: yeap correct. I thought we are talking about the banner :laughing:

Iamsamwhoami: hehe no

PR #30, opened 2019-04-04
"My charity" iterations

gustafguner:
The following is added to the charity screen:
* A charity logo image
* An info button
* A draggable overlay with charity information


Src/Classes/Ease.swift
```diff
@@ -0,0 +1,179 @@
+import Foundation
+import QuartzCore
+import Flow
+
+public final class Ease<T: Easeable> {
```
Iamsamwhoami: We wouldn't wanna leave these in our project, bring them out to our fork instead. `https://github.com/HedvigInsurance/ease`

Src/Classes/EaseDisposable.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+public typealias EaseDisposal = [EaseDisposable]
+
+public final class EaseDisposable {
```
Iamsamwhoami: I don't think this one is needed anymore.

Src/Generic/Extensions/UIImageView+Cache.swift
```diff
@@ -0,0 +1,33 @@
...
+//  Created by Gustaf Gunér on 2019-03-21.
+//  Copyright © 2019 Hedvig AB. All rights reserved.
+//
+
+import Foundation
+import UIKit
+import Disk
+
+extension UIImageView {
+    func imageFromURL(url: String) -> Void {
```
Iamsamwhoami: Not too sure if I like this, due to it being so hidden from the consumer (i.e a hidden function for initializing)
would prefer a viewable called `CachedImage` I think.

Src/Generic/Styling/TextStyle.swift
```diff
@@ -117,4 +122,9 @@ extension TextStyle {
         style.font = HedvigFonts.circularStdBold!.withSize(17)
         style.color = .purple
     }
+    
+    static let drabbableOverlayTitle = TextStyle.default.restyled { (style: inout TextStyle) in
```
Iamsamwhoami: Can we make the name more generic, maybe sync with Zak, he has an idea of what these are supposed to be called.

Src/Generic/Viewables/Button.swift
```diff
@@ -15,13 +15,16 @@ import UIKit
...
     case pillTransparent(backgroundColor: HedvigColor, textColor: HedvigColor)
+    case iconTransparent(textColor: HedvigColor, icon: ImageAsset?)
 
     func backgroundOpacity() -> CGFloat {
         switch self {
         case .standard:
             return 1
         case .pillTransparent:
             return 0.6
+        default:
```
Iamsamwhoami: Avoid using default here, use the `.iconTransparent` case instead, also applies to the other ones.

Src/Generic/Viewables/Button.swift
```diff
@@ -15,13 +15,16 @@ import UIKit
 enum ButtonType {
     case standard(backgroundColor: HedvigColor, textColor: HedvigColor)
     case pillTransparent(backgroundColor: HedvigColor, textColor: HedvigColor)
+    case iconTransparent(textColor: HedvigColor, icon: ImageAsset?)
```
Iamsamwhoami: `ImageAsset` should probably not be an Optional right, because that wouldn't be a valid case?

Src/Generic/Viewables/MarkdownText.swift
```diff
@@ -0,0 +1,83 @@
...
+//
+//  Created by Gustaf Gunér on 2019-04-04.
+//
+
+import Foundation
+import Flow
+import UIKit
+import MarkdownKit
+
+enum MarkdownTextStyle {
```
Iamsamwhoami: Could we reuse the `TextStyle` struct somehow?

Src/Screens/ProfileScreen/Screens/CharityScreen/Viewables/CharityInformationBody.swift
```diff
@@ -0,0 +1,36 @@
...
+//  CharityInformationBody.swift
+//  ugglan
+//
+//  Created by Gustaf Gunér on 2019-04-04.
+//
+
+import Foundation
+import Flow
+import UIKit
+import MarkdownKit
```
Iamsamwhoami: You probably don't need to import this here.

Src/Screens/ProfileScreen/Screens/CharityScreen/Viewables/CharityInformationBody.swift
```diff
@@ -0,0 +1,36 @@
...
+
+struct CharityInformationBody {
+    let text: String
```
Iamsamwhoami: I've been kinda leaning against always taking in `Signals`, up to you but it makes the components a bit more flexible.

Src/Screens/ProfileScreen/Screens/CharityScreen/Viewables/CharityLogo.swift
```diff
@@ -0,0 +1,39 @@
...
+//  Created by Gustaf Gunér on 2019-03-21.
+//
+
+
+import Flow
+import Foundation
+import UIKit
+
+struct CharityLogo {
+    let url: String
```
Iamsamwhoami: In Swift land a url is never a `String` but a `URL` 😉

Src/Screens/ProfileScreen/Screens/CharityScreen/Viewables/CharityLogo.swift
```diff
@@ -0,0 +1,39 @@
...
+    
+    init(url: String) {
```
Iamsamwhoami: This initializer is not needed.

Src/Generic/Viewables/Button.swift
gustafguner: Correct!

Src/Classes/Ease.swift
gustafguner: Whoops, completely forgot that this was inline 😅

Iamsamwhoami:
I saw some visual things when I looked at the built app also:

1. Draggable overlay can be dragged beyond the top of the screen, should stop right before overshooting I guess
2. The "handle" (little white bar) should have be white and a bit semi translucent.
3. Maybe we need to increase velocity a bit when swiping the draggable overlay, something doesn't feel 100% right 🤔
4. The button "Hur fungerar välgörenhet" should have some kind of highlight state.

Src/Generic/Styling/TextStyle.swift
gustafguner: Synced with Zak, turns out we're gonna have to change the look of the "header" of the draggable overlay a bit. I'll fix that.

Src/Generic/Extensions/UIImageView+Cache.swift
gustafguner: Good point!

Src/Screens/ProfileScreen/Screens/CharityScreen/Viewables/CharityLogo.swift
gustafguner: Swift land never ceases to amaze me 🙈

PR #31, opened 2019-04-16
Claims

Iamsamwhoami:

PR #32, opened 2019-04-29
Swift 5

Iamsamwhoami:
🎉


Src/Assets/HedvigColors.swift
```diff
@@ -55,6 +55,8 @@ extension UIColor {
             return .white
         case .offWhite:
             return .offWhite
+        case .yellow:
+            return .white
```
Iamsamwhoami: this will be fixed in #31

Src/Generic/Extensions/DeviceKit+RoundedCorners.swift
```diff
@@ -10,16 +10,16 @@ import DeviceKit
 import Foundation
 
 extension Device {
-    static var hasRoundedCorners: Bool = Device().isOneOf([
+    static var hasRoundedCorners: Bool = Device.current.isOneOf([
```
axelbacklund: DeviceKit for Swift 5 has the following: "New .allDevicesWithRoundedDisplayCorners and .hasRoundedDisplayCorners values to check if a device has rounded display corners. (eg. iPhone Xs and iPad Pro (3rd generation))"

PR #39, opened 2019-05-28
Tap feedback on perils

gustafguner:
Added haptic tap feedback on perils

### DIFF
--- a/Src/Generic/Extensions/UIControl+DelayedTouchCancel.swift
+++ b/Src/Generic/Extensions/UIControl+DelayedTouchCancel.swift
@@ -10,7 +10,7 @@ import Foundation
 import UIKit
 
 extension UIControl {
-    func delayedTouchCancel(delay: Double = 0.2) -> Signal<Bool> {
+    func delayedTouchCancel(delay: TimeInterval = 0.2) -> Signal<Void> {
         return Signal { callback in
             let bag = DisposeBag()
             
@@ -28,7 +28,7 @@ extension UIControl {
                 ).withLatestFrom(touchDownDateSignal.atOnce().plain())
                 .delay(by: { _, date in date.timeIntervalSinceNow < -delay ? 0 : delay })
                 .onValue({ _ in
-                    callback(true)
+                    callback(())
                 })
             
             return bag
--- a/Src/Generic/Extensions/UIControl+DelayedTouchCancel.swift
+++ b/Src/Generic/Extensions/UIControl+DelayedTouchCancel.swift
@@ -1,0 +1,37 @@
+//
+//  UIControl+DelayedTouchCancel.swift
+//  Ugglan
+//
+//  Created by Gustaf Gunér on 2019-05-29.
+//
+
+import Flow
+import Foundation
+import UIKit
+
+extension UIControl {
+    func delayedTouchCancel(delay: Double = 0.2) -> Signal<Bool> {
+        return Signal { callback in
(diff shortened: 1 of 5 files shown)


Src/Screens/DashboardScreen/Viewables/ExpandableRow.swift
```diff
@@ -56,10 +56,19 @@ extension ExpandableRow: Viewable {
...
 
         let expandableStackView = UIStackView()
+        expandableStackView.backgroundColor = .white
         expandableStackView.axis = .vertical
 
-        bag += expandableStackView.addArranged(content)
-
+        let contentWrapper = UIControl()
+        
+        bag += contentWrapper.add(content) { buttonView in
```
Iamsamwhoami: `contentWrapperView` maybe instead of `buttonView`

gustafguner: yeah that's more consistent 👍

Src/Screens/DashboardScreen/Viewables/Peril.swift
```diff
@@ -99,14 +99,27 @@ extension Peril: Reusable {
...
             }
             
-            bag += perilView.signal(for: .touchUpInside).animated(style: SpringAnimationStyle.lightBounce()) { _ in
-                perilView.transform = CGAffineTransform.identity
+            bag += merge(
+                perilView.signal(for: .touchUpInside),
+                perilView.signal(for: .touchUpOutside),
+                perilView.signal(for: .touchCancel)
+                ).withLatestFrom(touchDownDateSignal.atOnce().plain())
+                .delay(by: { _, date in date.timeIntervalSinceNow < -0.2 ? 0 : 0.2 })
```
Iamsamwhoami: maybe we should put this in an extension on `SignalProvider`, something like `delayedTouchDown` 🤔

gustafguner: hmm yeah that's true, definitely not the last time we're gonna need this 👌

gustafguner: but rather `delayedTouchCancel` I guess 🤔

Iamsamwhoami: Oh, yes! `delayedTouchCancel` is a good name!

Src/Generic/Extensions/UIControl+DelayedTouchCancel.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+//  Created by Gustaf Gunér on 2019-05-29.
+//
+
+import Flow
+import Foundation
+import UIKit
+
+extension UIControl {
+    func delayedTouchCancel(delay: Double = 0.2) -> Signal<Bool> {
```
Iamsamwhoami: This could be a `Signal<Void>` right? or does the Bool mean something?

Iamsamwhoami: Usually time is presented with the TimeInterval type, which is a typealias for a Double, but still 😄

gustafguner: Yep yep, you're right

gustafguner: The more descriptive the better 😄

PR #40, opened 2019-05-29
Fix draggable overlay title clipping

gustafguner:
Fixed the draggable overlay title clipping by increasing the line height and also added support for multiline titles


Iamsamwhoami:
@gustafguner Line height should be built in into form, have you tried that? https://github.com/iZettle/Form/blob/master/Form/TextStyle.swift#L71

PR #44, opened 2019-06-10
Referrals accept discount

Iamsamwhoami:


Src/Screens/ReferralsReceiverConsentScreen/Viewables/ReferralsReceiverConsentContent.swift
```diff
@@ -0,0 +1,129 @@
...
+//
+//  Created by Sam Pettersson on 2019-06-10.
+//
+
+import Foundation
+import Flow
+import UIKit
+import Form
+
+struct ReferralsReceiverConsentContent {
```
gustafguner: unique names are indeed a pain in the ass 😆

PR #55, opened 2019-07-12
Toasts 🍞

gustafguner:
A neat little addition to the app. I added this to a successful save of my info to demonstrate it. We could implement it in more places.

Skärmavbild 2019-07-12 kl  14 52 29

The toasts are quite customizable 💅 
- Symbol – Either a `String` (preferably an emoji) or an `Asset`
- Body – `String`
- Text color – `UIColor` (default: `offBlack`)
- Background color – `UIColor` (default: `white`)
- Duration – `TimeInterval` (default: `5.0`)

So for example

```swift
UIApplication.shared.appDelegate.createToast(
    symbol: .character("🤖"),
    body: "Hello world!",
    textColor: .white,
    backgroundColor: .turquoise,
    duration: 10.0
)
```

Would create this toast

Skärmavbild 2019-07-12 kl  15 12 11


Src/AppDelegate.swift
```diff
@@ -50,6 +68,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         ).onValue { _ in
             let loggedIn = LoggedIn()
             self.bag += self.window.present(loggedIn, options: [], animated: true)
+            
+            let toasts = Toasts(toastSignal: self.toastSignal)
```
axelbacklund: Maybe move everything here to its own method to make it a bit cleaner?

Src/Toasts.swift
```diff
@@ -0,0 +1,218 @@
...
+                make.width.equalTo(20)
+            }
+            
+            return symbol
+        }
+        
+        return UIView()
+    }
+    
+    func getContainerEdgeInsets() -> UIEdgeInsets {
```
Iamsamwhoami: maybe a computed property instead?

Src/Toasts.swift
```diff
@@ -0,0 +1,218 @@
...
+import Flow
+import Form
+import UIKit
+
+enum ToastHideDirection {
+    case left
+    case right
+}
+
+enum ToastSymbol {
```
Iamsamwhoami: You could probably use an `Either<ImageAsset, Character>` instead

Src/Generic/Extensions/UIView/UIView+Viewable.swift
```diff
@@ -142,16 +142,17 @@ extension UIView {
 extension UIStackView {
     // swiftlint:disable large_tuple
     private func materializeArrangedViewable<V: Viewable, MatterView: UIView>(
-        viewable: V
+        viewable: V,
+        atIndex: Int?
```
Iamsamwhoami: As this is not used remove it maybe?

gustafguner: Yeah, not using anymore 👍

Src/Toasts.swift
gustafguner: Neat!

PR #75, opened 2020-01-29
dynamic background color added

Pavelbarrosq:


Src/Screens/ReferralsScreen/Viewables/ReferralsProgressHighPremium.swift
```diff
@@ -24,7 +24,7 @@ extension ReferralsProgressHighPremium: Viewable {
         view.layoutMargins = UIEdgeInsets(top: 0, left: 16, bottom: 10, right: 16)
 
         let backgroundView = UIView()
-        backgroundView.backgroundColor = .turquoise
+        backgroundView.backgroundColor = .dynamicBackgroundColor
```
sampettersson: As this is very specific for this view, move the color directly into here instead I would say:

```swift
UIColor(dynamic: { trait -> UIColor in
trait.userInterfaceStyle == .dark ? .grass700 : .turquoise
})
```

PR #80, opened 2020-02-03
KeyGear valuation date picker screen added

Pavelbarrosq:


Src/Assets/HedvigColors.swift
```diff
@@ -31,9 +31,16 @@ extension UIColor {
         })
     }
     
+    static var switchPrimaryText: UIColor {
```
sampettersson: primaryTextNeg?

Src/Generic/Styling/TextStyle.swift
```diff
@@ -254,4 +254,376 @@ extension TextStyle {
...
+        style.color = .primaryText
+    }
+    
+    static let headlineLargeNegLargeNegLeft = TextStyle.default.restyled { (style: inout TextStyle) in
+        style.font = HedvigFonts.soRayExtraBold!.withSize(24)
+        style.lineHeight = 24
+        style.color = .switchPrimaryText
+    }
+    
+    static let headlineLargeLargCenter = TextStyle.default.restyled { (style: inout TextStyle) in
```
sampettersson: I know this is wrong in zeplin too, but it should be `LargeLarge` not `LargeLarg` I think 😄

PR #88, opened 2020-03-04
Presentables for draggableOverlay fix

Pavelbarrosq:


Src/Screens/OfferScreen/ApplyDiscount.swift
```diff
@@ -26,38 +26,43 @@ struct ApplyDiscount {
 extension ApplyDiscount: Presentable {
     func materialize() -> (UIViewController, Future<Void>) {
         let viewController = UIViewController()
+        viewController.preferredContentSize = CGSize(width: 0, height: 0)
```
sampettersson: doesn't the animation screw up when you do this? 🤔

Pavelbarrosq: it dosen't make any difference if i have it or not 😅The animation works good either way but i erase that line.

PR #89, opened 2020-03-04
Content insurance card created

Pavelbarrosq:


Src/Viewables/ContentInsuranceCard.swift
```diff
@@ -0,0 +1,83 @@
...
+import Flow
+import Form
+import Presentation
+import UIKit
+
+enum ContentAction {
+    case toContentInsurance, toTravelInsurance
+}
+
+struct ContentInsuranceCard {
```
sampettersson: Maybe call this something more generic? like card with image, text and action.

Src/Viewables/ContentInsuranceCard.swift
```diff
@@ -0,0 +1,83 @@
...
+    let iconImage: ImageAsset
+    let title: String
+    let description: String
+    let type: ContentAction
+    let presentingViewController: UIViewController
+}
+
+extension ContentInsuranceCard: Viewable {
+    func materialize(events: ViewableEvents) -> (UIView, Future<Void>) {
```
sampettersson: Make it into a signal instead, and then the consumer will have the responsibility for showing the subscreen.

Src/Viewables/ContentInsuranceCard.swift
```diff
@@ -0,0 +1,83 @@
...
+import Presentation
+import UIKit
+
+enum ContentAction {
+    case toContentInsurance, toTravelInsurance
+}
+
+struct ContentInsuranceCard {
+    let iconImage: ImageAsset
+    let title: String
```
sampettersson: make these into signals so they can be changed dynamically.

Src/Viewables/ContentInsuranceCard.swift
```diff
@@ -0,0 +1,83 @@
...
+        stackView.axis = .vertical
+        stackView.alignment = .center
+        stackView.edgeInsets = UIEdgeInsets(top: 16, left: 15, bottom: 24, right: 15)
+        
+        let icon = Icon(icon: self.iconImage, iconWidth: 40)
+        stackView.addArrangedSubview(icon)
+        
+        bag += stackView.addArranged(Spacing(height: 16))
+        
+        let titleLabel = MultilineLabel(value: "", style: .headlineSmallSmallLeft)
```
sampettersson: you should pass in the title here.

PR #96, opened 2020-03-30
Redesigned chat + added dynamic colors to Hedvigcolors

Pavelbarrosq:


Src/Assets/HedvigColors.swift
```diff
@@ -100,6 +100,24 @@ extension UIColor {
...
+        })
+    }
+    
+    static var boxPrimaryBackground: UIColor {
+        UIColor(dynamic: { trait -> UIColor in
+            trait.userInterfaceStyle == .dark ? .darkerGray : .lighterGray
+        })
+    }
+    
+    static var alternativeBackground: UIColor {
```
sampettersson: tertiary background maybe? we already have primary and secondary.

Src/Screens/ChatScreen/Reusables/Message+Reusable.swift
```diff
@@ -458,11 +458,22 @@ extension Message: Reusable {
...
+                            style: TextStyle.bodyRegularRegularLeft
+                        )
+                        bag += contentContainer.addArranged(label)
+                    } else {
+                        let label = MultilineLabel(
+                            value: message.body,
+                            style: TextStyle.bodyRegularNegRegularNegLeft
+                        )
+                        bag += contentContainer.addArranged(label) { label in
+                            label.textColor = .black
```
sampettersson: don't use textColor directly, use .colored on the text style instead.

PR #139, opened 2021-02-02
[HVG-869] Embark UI Updates

tarikstafford:


Projects/Embark/Sources/Viewables/Actions/TextActionSet.swift
```diff
@@ -35,18 +46,28 @@ extension TextActionSet: Viewable {
                 shouldAutoFocus: index == 0,
                 fieldStyle: .embarkInputSmall
             )
-            return (signal: view.addArranged(input), shouldReturn: input.shouldReturn, action: textAction)
+            
+            let isFirstName: Bool = textAction.data?.key == "firstName"
```
sampettersson, line 50: You shouldn't use a hard-coded key value to identify this behaviour, what is the effect you want to achieve here? :)

Projects/Embark/Sources/Viewables/Actions/TextActionSet.swift
```diff
@@ -61,6 +82,11 @@ extension TextActionSet: Viewable {
                     callback(link.fragments.embarkLinkFragment)
                 }
             }
+            
+            containerView.addSubview(boxStack)
+            boxStack.snp.makeConstraints({ $0.edges.equalToSuperview() })
```
sampettersson, line 87: NIT: prefer `make in`

Projects/Embark/Sources/Viewables/Actions/TextActionSet.swift
```diff
@@ -72,6 +98,36 @@ extension TextActionSet: Viewable {
...
             }
+            
+            bag += containerView.applyShadow({ (traitCollection) -> UIView.ShadowProperties in
+                UIView.ShadowProperties(
+                    opacity: 1,
+                    offset: .init(width: 0, height: 2),
+                    radius: 2,
+                    color: .brand(.secondaryShadowColor),
+                    path: nil,
+                    corners: (.allCorners, 8))
```
sampettersson, line 109: Unsure what I feel about a tuple here, maybe make a struct holding this information, like CornerStyle or something, and then you can do `.defaultAllCorners` or something :)

Projects/Embark/Sources/Viewables/Actions/TextActionSet.swift
```diff
@@ -72,6 +98,36 @@ extension TextActionSet: Viewable {
...
+
+            bag += view.addArranged(button)
+            
+            func isValid(signal: ReadWriteSignal<String>, action: TextAction) -> Signal<Bool> {
+                signal.map { text in
+                    return text.count > 0
+                }.plain()
+            }
+            
+            bag += combineLatest(textActions!.map { signal, _, action in isValid(signal: signal, action: action) }).map {
```
sampettersson, line 126: Avoid unwrapping here

sampettersson: Maybe you could make this a bit cleaner with allSatisfy?

Projects/hCore/Sources/UIView+DynamicShadow.swift
```diff
@@ -9,45 +9,52 @@ public extension UIView {
...
+                        }
+                        
+                        if let radius = properties.radius {
+                            self.layer.shadowRadius = radius
+                        }
+                        
+                        if let corners = properties.corners {
+                            self.layer.shadowPath = UIBezierPath(roundedRect: self.bounds, byRoundingCorners: corners.corner, cornerRadii: CGSize(width: corners.radius, height: corners.radius)).cgPath
+                        }
+                        self.layer.shouldRasterize = true
```
sampettersson, line 55: Does rasterisation have any rendering impacts? I'm not sure we should always force it to rasterise or?

Projects/hCoreUI/Sources/HedvigColors.swift
```diff
@@ -132,6 +139,18 @@ public extension UIColor {
                 return BrandColorBase.link
             case .destructive:
                 return BrandColorBase.destructive
+            case .secondaryShadowColor:
+                return UIColor(red: 0, green: 0, blue: 0, alpha: 0.1)
+            case .embarkMessageBubble(let negative):
```
sampettersson, line 144: NIT: I usually prefer `case let` over `.value(let variable)`

sampettersson (changes requested):

Projects/Embark/Sources/Viewables/Actions/TextActionSet.swift
tarikstafford: True. I think the thing that's confusing is that radius here is actually not the radius but the spread of the shadow. maybe renaming that to something that makes more sense.

Projects/hCore/Sources/UIView+DynamicShadow.swift
tarikstafford: It's probably only necessary in scroll views / animated views. I think in static views its pointless? Let's make it a shadow property.

Projects/hCoreUI/Sources/HedvigColors.swift
tarikstafford: HA TIL!

Projects/Embark/Sources/Viewables/Actions/TextActionSet.swift
tarikstafford: this is before I knew how embark worked!

Projects/Embark/Sources/Viewables/Actions/TextActionSet.swift
tarikstafford: That's interesting can you give me an example?

Projects/Embark/Sources/Viewables/Actions/TextActionSet.swift
```diff
@@ -35,18 +46,28 @@ extension TextActionSet: Viewable {
                 shouldAutoFocus: index == 0,
                 fieldStyle: .embarkInputSmall
             )
-            return (signal: view.addArranged(input), shouldReturn: input.shouldReturn, action: textAction)
+            
+            let isFirstName: Bool = index == 0
```
sampettersson, line 50: Isn't what you should do here is instead `isLastTextAction` and !isLastTextAction render a divider? :) I think text action can have any number of inputs, so there might be cases where we have 3 instead of two items.

Projects/Embark/Sources/Viewables/Actions/TextActionSet.swift
```diff
@@ -35,18 +46,28 @@ extension TextActionSet: Viewable {
...
+            let stack = UIStackView()
+            stack.axis = .horizontal
+            stack.distribution = .equalSpacing
+            
+            stack.addArrangedSubview(label)
+            
+            boxStack.addArrangedSubview(stack)
+            
+            if isFirstName {
+                let view = UIView.init(height: 0.5)
```
sampettersson, line 63: There is a `Divider` component 😄

Projects/Embark/Sources/Viewables/Embark+ShadowProperties.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+import UIKit
+import hCore
+
+internal extension UIView.ShadowProperties {
+    static let embark = UIView.ShadowProperties(
```
sampettersson, line 6: Noice!

sampettersson (changes requested):

sampettersson (approved):

PR #145, opened 2021-02-10
Hvg 893 embark tracking

tarikstafford:


sampettersson (dismissed):
Neat! Code looks good, maybe we should write a test for this logic though?

Projects/Embark/Sources/EmbarkState+Tracking.swift
```diff
@@ -0,0 +1,58 @@
...
+}
+
+public extension EmbarkTrackingEvent {
+    static var trackingHandler: (_ Event: EmbarkTrackingEvent) -> Void = { _ in }
+}
+
+internal extension GraphQL.EmbarkExternalRedirectLocation {
+    func trackingEvent(store: EmbarkStore) -> EmbarkTrackingEvent {
+        var properties = [String:String]()
+        if let store = store.revisions.last {
```
sampettersson, line 21: maybe we should have a `func getAllValues() -> [String: String]` and that one just goes into revisions and picks the last one. what you think?

sampettersson, line 21: Use `getAllValues()` here too?

Projects/Embark/Sources/EmbarkStore.swift
```diff
@@ -36,6 +36,10 @@ class EmbarkStore {
             }
         }
     }
+    
+    func getAllValues() -> [String:String] {
+        return revisions.last ?? [:]
```
sampettersson, line 41: We would need to evaluate all computed values here before returning right? Also will we get any performance implications by that 🤔

sampettersson (changes requested):

Projects/Embark/Sources/EmbarkState.swift
```diff
@@ -50,8 +54,18 @@ public struct EmbarkState {
             return computedValues
         } ?? [:]
     }
+    
+    func startTracking() {
+        bag += currentPassageSignal
+                .readOnly()
+                .compactMap { $0?.tracks }
+                .onValue({ (tracks) in
```
sampettersson, line 62: Use the background scheduler here: `on: .background`

Projects/Embark/Sources/EmbarkState+Tracking.swift
```diff
@@ -0,0 +1,55 @@
...
+
+internal extension EmbarkPassage.Track {
+    func trackingEvent(store: EmbarkStore) -> EmbarkTrackingEvent {
+        return .init(title: self.eventName, properties: properties(store: store))
+    }
+    
+    private func properties(store: EmbarkStore) -> [String:Any] {
+        var properties = Dictionary<String,Any>()
+        if includeAllKeys {
+            properties = properties.merging((store.getAllValues())) { return $1 }
```
sampettersson, line 34: You can use `takeRight` here.

Projects/Embark/Sources/EmbarkState+Tracking.swift
```diff
@@ -0,0 +1,55 @@
...
+        } else {
+            let storeProperties = store.getAllValues().filter { key, value in
+                return eventKeys.contains(key)
+            }
+            
+            properties = properties.merging(storeProperties, uniquingKeysWith: { return $1 })
```
sampettersson, line 40: You can use `takeRight` here.

Projects/Embark/Sources/EmbarkState+Tracking.swift
```diff
@@ -0,0 +1,55 @@
...
+        }
+        
+        if let customData = customData {
+            properties = properties.merging((customData.toJSONDictionary() ?? [:]), uniquingKeysWith: { return $1 })
```
sampettersson, line 44: You can use `takeRight` here.

sampettersson (changes requested):

Projects/Embark/Sources/EmbarkState+Tracking.swift
```diff
@@ -0,0 +1,55 @@
...
+}
+
+internal extension EmbarkPassage.Track {
+    func trackingEvent(store: EmbarkStore) -> EmbarkTrackingEvent {
+        return .init(title: self.eventName, properties: properties(store: store))
+    }
+    
+    private func properties(store: EmbarkStore) -> [String:Any] {
+        var properties = Dictionary<String,Any>()
+        if includeAllKeys {
```
sampettersson, line 33: it's been decided that this is not something we wanna support going forward, due to there being unwanted data sent to Mixpanel, so we can remove this for now.

Projects/Embark/Sources/EmbarkState+Tracking.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+public extension EmbarkTrackingEvent {
+    static var trackingHandler: (_ Event: EmbarkTrackingEvent) -> Void = { _ in }
+}
+
+internal extension GraphQL.EmbarkExternalRedirectLocation {
+    func trackingEvent(store: EmbarkStore) -> EmbarkTrackingEvent {
+        var properties = store.getAllValues()
```
sampettersson, line 20: Don't send the whole store here.

Projects/Embark/Sources/EmbarkState+Tracking.swift
tarikstafford: But we should still send the entire dictionary on external redirect ?

sampettersson (approved):
You probably need to merge this against `embark-feature` for the build to run.

PR #146, opened 2021-02-11
[HVG-884] Embark computed values

sampettersson:


Projects/Embark/Sources/ComputedValues/Array<Token>+AbstractExpression.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+
+indirect enum Expression: Equatable {
```
tarikstafford, line 3: very cool

tarikstafford (approved):

PR #163, opened 2021-03-31
App 284 add markdown support for messages

tarikstafford:


Projects/hCoreUI/Sources/MarkdownText.swift
```diff
@@ -53,6 +54,8 @@ extension MarkdownText: Viewable {
             }
         }
 
+        let tapGesture = UITapGestureRecognizer()
```
sampettersson, line 57:
```suggestion
```

Projects/hCoreUI/Sources/MarkdownText.swift
```diff
@@ -41,6 +41,7 @@ extension MarkdownText: Viewable {
         markdownText.numberOfLines = 0
         markdownText.lineBreakMode = .byWordWrapping
         markdownText.baselineAdjustment = .none
+        markdownText.isUserInteractionEnabled = true
```
sampettersson, line 44:
```suggestion
```

sampettersson (changes requested):
Good find on using a UITextView instead ⛱️

Projects/hCoreUI/Sources/MarkdownText.swift
tarikstafford: I think you need this for links to be tappable.

sampettersson (approved):

PR #164, opened 2021-03-31
Embark fixes

sampettersson:
See commits for what 😇

### DIFF
--- a/Projects/Embark/Sources/Viewables/Actions/Masking.swift
+++ b/Projects/Embark/Sources/Viewables/Actions/Masking.swift
@@ -17,10 +17,10 @@ struct Masking {
     func isValid(text: String) -> Bool {
         switch type {
         case .personalNumber:
-            let age = age(text: text) ?? 0
+            let age = calculateAge(from: text) ?? 0
             return text.count > 10 && 15 ... 130 ~= age
         case .birthDate, .birthDateReverse:
-            let age = age(text: text) ?? 0
+            let age = calculateAge(from: text) ?? 0
             return 15 ... 130 ~= age
         case .email:
             let emailRegEx = "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}"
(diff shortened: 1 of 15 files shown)


Projects/Embark/Sources/Viewables/Actions/TextActionSet.swift
```diff
@@ -37,19 +37,20 @@ extension TextActionSet: Viewable {
...
             }
 
+            let endIndex = (data.textActionSetData?.textActions.endIndex ?? 1)
+            let isLastAction: Bool = index == endIndex - 1
+
             let input = EmbarkInput(
                 placeholder: textAction.data?.placeholder ?? "",
                 keyboardType: masking?.keyboardType,
                 textContentType: masking?.textContentType,
+                returnKeyType: isLastAction ? .done : .next,
```
tarikstafford, line 49: nice add

Projects/Embark/Sources/Viewables/Actions/Masking.swift
```diff
@@ -14,8 +14,25 @@ enum MaskType: String {
...
     let type: MaskType
 
-    func isValid(text _: String) -> String {
-        ""
+    func isValid(text: String) -> Bool {
+        switch type {
+        case .personalNumber:
+            return text.count > 10
+        case .birthDate, .birthDateReverse:
+            return text.count == 10
```
tarikstafford, line 22: if the formatting is slightly different from be won't this fail? I saw on the Moving Flow - birthdate was formatted without dashes. Don't we get the max min from BE for validation?

tarikstafford (approved):
Looks good, just concerned about birthdate validation!

Projects/Embark/Sources/Viewables/Actions/Masking.swift
sampettersson: The function here makes sure the formatting is always correct: https://github.com/HedvigInsurance/ugglan/blob/e967efe69cb87d508897852681ae6172ed28c75a/Projects/Embark/Sources/Viewables/Actions/Masking.swift#L176

I am guessing in the case of moving flow a mask hasn't been used/set at all in angel, but in that case this function wouldn't be called at all 😄 We should also implement min/max though.

tarikstafford (approved):

PR #166, opened 2021-04-07
APP-85-native-simple-sign

tarikstafford:


sampettersson:
@tarikstafford tried running this locally but it Ugglan doesn't compile 😢

Projects/hCoreUI/Sources/ButtonRow.swift
```diff
@@ -70,12 +70,14 @@ public struct ButtonRowViewWrapper {
     public let onTapSignal: Signal<Void>
     public let type: ButtonType
     public let animate: Bool
-    public let isEnabled: Bool
+    private let isEnabled: Bool
```
sampettersson, line 73: I guess remove this if we use it as a signal instead?

Projects/App/Sources/Generic/AppFlow.swift
```diff
@@ -6,27 +6,34 @@ import hCoreUI
...
     let window: UIWindow
+    let bag = DisposeBag()
 
     init(window: UIWindow) {
         self.window = window
         self.window.rootViewController = rootNavigationController
     }
+
+    func presentLoggedIn() {
+        ApplicationState.preserveState(.loggedIn)
```
sampettersson, line 21: Isn't this done inside `LoggedIn`?

Projects/App/Sources/Generic/AppFlow.swift
```diff
@@ -95,9 +102,11 @@ extension EmbarkOnboardingFlow {
         case .login:
             return Presentation(
                 Login(),
-                style: .detented(.medium, .large),
-                options: [.allowSwipeDismissAlways, .defaults]
-            )
+                style: .detented(.large),
```
sampettersson, line 105: it should be `.detented(.medium)` for Sweden. I guess we should use `preferredPresentationStyle` and just `.default`.

sampettersson (changes requested):

Projects/App/Sources/Screens/BankIDLoginScreen/SimpleSignLogin.swift
```diff
@@ -0,0 +1,117 @@
...
+        let form = FormView()
+
+        let masking = Localization.Locale.currentLocale.market.masking
+
+        viewController.title = L10n.bankidLoginTitle
+
+        let label = UILabel()
+        label.text = Localization.Locale.currentLocale.market.labelTitle
+
+        let textField = UITextField()
```
sampettersson, line 29: What do you think about making a utility func on masking that applies it to a `UITextField`, something like `Masking.applyTo(_ textField: UITextField) -> Bag` which handles `isValid`, `textContentType`, `keyboardType`, etc.

tarikstafford: How would you add `isValid` in this case? Perhaps it needs to be thought out a bit more before we add a convenience function for it.

tarikstafford: I guess `maskValue` would make sense, although I am not really using this. Actually it's never really masking the value is it? Just formatting it.

tarikstafford: I tried doing this a few ways, but I don't really like the idea of a method returning a signal for validity in a masking object. Do you think we should split up `masking` and `validation` -> perhaps we can change `maskType` to something like `textFieldType` or `textFieldContentType` ?

tarikstafford: Also the textField settings would be configurable from the type itself perhaps? Then masking would just have an `applyMask` and `validation` would have a `isValidSignal`

sampettersson: @tarikstafford don't we always wanna mask? it makes sure that the user can only enter the valid characters, also it inserts the hyphens at the correct places and so forth also we have the same predictable behaviour in all places of the app. And then we can just use `unmaskValue` before sending to backend.

Projects/hCore/Sources/Masking.swift
```diff
@@ -16,10 +17,34 @@ public enum MaskType: String {
...
+    public func isValidSignal(_ textField: UITextField) -> ReadSignal<Bool> {
+        textField.distinct().map { text in isValid(text: text) }
+    }
+
+    public func applyMasking(_ textField: UITextField) -> Disposable {
+        let bag = DisposeBag()
+
+        bag += textField.distinct().onValue { text in
+            let newValue = maskValue(text: text, previousText: previousText)
+            $previousText.value = newValue
```
sampettersson, line 41: Way nicer than before! 😄

sampettersson (approved):

sampettersson (approved):

PR #186, opened 2021-04-12
[APP-320] Width constrain actions

sampettersson:


Projects/Embark/Sources/Viewables/Actions/Action.swift
```diff
@@ -22,11 +22,26 @@ struct ActionResponseData {
...
+        outerContainer.axis = .vertical
+        outerContainer.alignment = .center
+        
+        let widthContainer = UIStackView()
+        widthContainer.axis = .horizontal
+        outerContainer.addArrangedSubview(widthContainer)
+        
+        bag += outerContainer.didLayoutSignal.onValue{ _ in
+            widthContainer.snp.remakeConstraints { make in
+                make.width.equalTo(outerContainer.frame.width > 600 ? 600 : outerContainer.frame.width)
```
tarikstafford, line 37: is this for ipad?

sampettersson: yep! was battling this for an insane while and this is the best solution I came up with 😒

tarikstafford: What about `UIDevice.current.userInterfaceIdiom == .pad` instead just so its clear what this is for?

tarikstafford (approved):

PR #189, opened 2021-04-13
Add pull request template

sampettersson:
To keep us on our toes, do you think we should add something more to the checklist? Tests is an obvious one, after the pipeline is working again that is 😆

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,10 @@
+## [APP-XXX]
+
+- Change 1
+- Change 2
+
+## Checklist
+
+- [ ] Dark/light mode verification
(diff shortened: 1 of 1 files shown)


tarikstafford (approved):

tarikstafford:
> To keep us on our toes, do you think we should add something more to the checklist? Tests is an obvious one, after the pipeline is working again that is 😆

Great idea! Nothing that comes to mind. Previously I used to write test notes for QA - and then if small screen size was considered - also if unit tests or ui tests are required.

sampettersson:
@tarikstafford smaller screens is a good one, added that too!

tarikstafford (approved):

PR #201, opened 2021-04-14
Fix auto responses being presented even though they contain no info

sampettersson:
## [APP-349]

- Fix auto responses being presented even though they contain no info

## Checklist

- [x] Landscape verification
- [x] iPad verification

[APP-349]: https://hedvig.atlassian.net/browse/APP-349


Projects/Embark/Sources/Viewables/EmbarkMessages.swift
```diff
@@ -100,20 +100,29 @@ extension EmbarkMessages: Viewable {
 
         let animatedResponseSignal: Signal = messagesDataSignal.withLatestFrom(previousResponseSignal).animated(style: .lightBounce(), animations: { _, previousResponse in
             if self.state.animationDirectionSignal.value == .forwards {
```
tarikstafford, line 102: Maybe make this a `guard` statement instead - so we don't nest IF statements?

tarikstafford (approved):

PR #225, opened 2021-05-28
App 328 multi action component

tarikstafford:
## [APP-328]

- Add multi-action 
- Need to update grouped response to handle <Each>

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

## Things to do

- [ ] Add snapshot tests
- [ ] Use store to cache display values
- [ ] Make sure all data is lined up properly
- [ ] Figure out a solution for when a user does not add any multi action objects?

[APP-328]: https://hedvig.atlassian.net/browse/APP-328


sampettersson (changes requested):
Code looks really good, some UI things I've noticed:

- [x] Box after saving always says `2kvadratmeter` regardless of what I put
- [x] I need to switch on and off before being able to click `Save`
- [ ] Add an animation when changing the height of the components so that it doesn't `PLOPP`
- [ ] Missing label for size
- [ ] Label instead of text is used in picker says Sauna instead of Bastu
- [ ] Detented modal doesn't adjust for keyboard popping up
- [ ] The gray BG on the section looks a bit off

sampettersson (approved):

PR #235, opened 2021-06-28
[APP-530] last quote of member fix

sampettersson:
## [APP-530]

Fetch offer id from `Query.lastQuoteOfMember` in Swedish OnboardingChat

## Checklist

- [x] Landscape verification
- [x] iPad verification

[APP-530]: https://hedvig.atlassian.net/browse/APP-530

### DIFF
--- a/Projects/App/Sources/Screens/ChatScreen/Chat.swift
+++ b/Projects/App/Sources/Screens/ChatScreen/Chat.swift
@@ -6,6 +6,7 @@ import Offer
 import Presentation
 import UIKit
 import hCore
+import hGraphQL
 
 struct Chat {
 	@Inject var client: ApolloClient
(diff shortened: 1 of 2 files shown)


Projects/App/Sources/Screens/ChatScreen/Chat.swift
```diff
@@ -46,9 +47,28 @@ extension Chat: Presentable {
...
 			case .offer:
-				viewController.present(
-					Offer(offerIDContainer: .stored, menu: Menu(title: nil, children: []))
-				)
+                client.fetch(query: GraphQL.LastQuoteOfMemberQuery()).onValue { data in
+                    guard let id = data.lastQuoteOfMember.asCompleteQuote?.id else {
+                        return
+                    }
+                    
+                    viewController.present(
```
tarikstafford, line 55: it would be great to pull navigation out of screen level - now that we have an offer module would it be possible to create a `ChatFlow` object that manages more of the navigation? Is this something we wanna commit to longterm - as in slowly moving `viewController.present` to a dedicated space so we don't run into cross dependency issues? ie we don't want Chat to import Offer?

tarikstafford (changes requested):

Projects/App/Sources/Screens/ChatScreen/Chat.swift
sampettersson: @tarikstafford yeah, get ya, `Chat` is a part of `App` now though, so lets do this later? this is gonna be obsolete quite soon anyways.

PR #262, opened 2021-07-05
Approve only button

sampettersson:
## [APP-539]

- Fix approve only button not showing

## Checklist

- [x] Landscape verification
- [x] iPad verification

[APP-539]: https://hedvig.atlassian.net/browse/APP-539


Projects/Offer/Sources/Offer.swift
```diff
@@ -209,7 +209,11 @@ class OfferState {
 		return client.perform(mutation: GraphQL.SignQuotesMutation(ids: ids))
 			.map { data in
 				if data.signQuotes.asAlreadyCompleted != nil {
-					self.$hasSignedQuotes.value = true
+					Future().delay(by: 1)
```
tarikstafford, line 212: What's the delay for here?

tarikstafford (changes requested):

Projects/Offer/Sources/Offer.swift
sampettersson: Because if signing was too fast it felt like something went wrong 😅

sampettersson: Buut, thinking about it signing in prod might be slow enough already 🤔

tarikstafford: Can't this introduce race conditions tho?

tarikstafford: Yeah I don't think we should introduce that much delay tho if we don't need it. Maybe 0.5 is ok.

tarikstafford (approved):

tarikstafford (approved):

PR #316, opened 2021-09-06
Fix market picker flaky

sampettersson:
> Moving over to use the MarketStore fully for the Market Picker screen, hopefully this makes the screen less flaky.


tarikstafford (approved):
looks good!

PR #326, opened 2021-09-13
Choose plan swiftui

sampettersson:
We had a bug where the discount pill wasn't sizing correctly, but then the design in Figma was updated anyways, so just rewrote parts of the ChoosePlanRow to SwiftUI.


tarikstafford (commented):
Screenshot 2021-09-15 at 10 57 41

PR #334, opened 2021-09-23
App 779 refactor features loader

tarikstafford:
## [APP-779]

refactor placeholder view controller -> window can be nil on viewDidLoad so we need to use viewDidAppear. It's still an ugly-ish solution though.

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

[APP-779]: https://hedvig.atlassian.net/browse/APP-779


sampettersson (approved):
Good enough for now I think 😄

PR #393, opened 2021-11-14
OTP Authentication

sampettersson:
## [APP-1067]

> Auth with email OTP for France

## Checklist

- [x] Landscape verification
- [x] iPad verification

[APP-1067]: https://hedvig.atlassian.net/browse/APP-1067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tarikstafford (changes requested):
Can you add test notes? How do I generate an email with an OTP?

Projects/Authentication/Sources/AuthenticationStore.swift
```diff
@@ -0,0 +1,215 @@
...
+import hCoreUI
+import hGraphQL
+
+struct OTPState: StateProtocol {
+    var isLoading = false
+    var isResending = false
+    var id: String? = nil
+    var code: String = ""
+    var codeErrorMessage: String? = nil
+    var emailErrorMessage: String? = nil
```
tarikstafford, line 16: What do you think about adding a default `StateProtocol` property called `var errors: (Error) -> Void` or something of that nature? Then when we fetch anything we can `.onError` and return a default `setError` ?

tarikstafford: Ah wait I forgot properties in `State` have to be codable, equitable -> maybe var errors: [Error]

Projects/Authentication/Sources/AuthenticationStore.swift
```diff
@@ -0,0 +1,215 @@
...
+                    )
+                )
+                .delay(by: 0.5)
+                .compactMap { data in
+                    if data.loginVerifyOtpAttempt.asVerifyOtpLoginAttemptError != nil {
+                        return .otpStateAction(
+                            action: .setCodeError(message: L10n.Login.CodeInput.ErrorMsg.codeNotValid)
+                        )
+                    } else if let success = data.loginVerifyOtpAttempt.asVerifyOtpLoginAttemptSuccess {
+                        return .navigationAction(action: .authSuccess(accessToken: success.accessToken))
```
tarikstafford, line 88: I wonder if we can start to disambiguate action types and have separate handlers for effects / reduction. Would that be an interesting thing to look into? I recall we thought about it earlier - do you think it would have value?

Projects/Authentication/Sources/OTP/OTPCodeDisplay.swift
```diff
@@ -0,0 +1,81 @@
+import Foundation
+import SwiftUI
+import hCore
+import hCoreUI
+
+struct ShakeEffect: GeometryEffect {
```
tarikstafford, line 6: is there a method to add a geo effect on a view extension so we don't have to call `.modifier` all the time

Projects/Authentication/Sources/OTP/ReadOTPState.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+import SwiftUI
+import hCore
+
+struct ReadOTPState<Content: View>: View {
```
tarikstafford, line 5: COOL! Can we just make this generic for all Stores?

Projects/hCore/Sources/PasteView.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    required init?(
+        coder: NSCoder
+    ) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func canPerformAction(_ action: Selector, withSender sender: Any?) -> Bool {
+        switch action {
```
tarikstafford, line 22: I have never seen this!

tarikstafford (approved):
Amazing code!

PR #415, opened 2021-12-06
Fix appearance not being applied

sampettersson:


tarikstafford (approved):

Projects/App/Sources/AppDelegate.swift
```diff
@@ -381,7 +381,6 @@ let log = Logger.builder
 
         launchView.snp.makeConstraints { make in make.top.bottom.leading.trailing.equalToSuperview() }
 
-        customNavigationController = { _ in NavigationController() }
```
tarikstafford, line 384: If you remove this it will break again no?

PR #418, opened 2021-12-08
Add Moving Flow to NO

sampettersson:
## [APP-1130]

> Add moving flow to NO, also changed ContractDetail to be resolved by its id so that we get live updates to the views as the dataset changes :)

## Checklist

- [x] Landscape verification
- [x] iPad verification

[APP-1130]: https://hedvig.atlassian.net/browse/APP-1130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Projects/App/Sources/UgglanStore.swift
```diff
@@ -47,6 +47,9 @@ public final class UgglanStore: StateStore<UgglanState, UgglanAction> {
                     query: GraphQL.FeaturesQuery(),
                     cachePolicy: .fetchIgnoringCacheData
                 )
+                .mapError({ error in
+                    .init(member: .init(features: []))
```
sampettersson, line 51: This caused an endless spinenr when this query was failing, so I guess its good to fallback to empty.

PR #427, opened 2021-12-14
trigger action before dismissing

tarikstafford:
Fix claim status polling never stopping

- Add new `sendActionOnDismiss`

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sampettersson:
Why do we want to stop polling?

sampettersson (approved):

PR #430, opened 2021-12-15
Payment at sign

sampettersson:
## [APP-1157]

> Move payment connection to before sign in DK and NO

Also code scouted and removed some `.onValue { _ in }` and replaced with a new `Future.sink` method.

## Checklist

- [x] Landscape verification
- [x] iPad verification

[APP-1157]: https://hedvig.atlassian.net/browse/APP-1157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Projects/hCore/Sources/ScrollView+Inspect.swift
```diff
@@ -33,17 +33,29 @@ struct ViewIntrospector<ViewType: UIView>: UIViewRepresentable {
...
+
+        guard let superview = view.superview else {
+            return nil
+        }
+
+        if let view = findView(from: superview) {
+            return view
+        }
+
+        return traverseUp(from: superview, levels: levels + 1)
```
tarikstafford, line 49: Is there a reason to only go 5 levels up? Can't SwiftUI have a lot of stacked views?

tarikstafford (approved):

Projects/hCore/Sources/ScrollView+Inspect.swift
sampettersson: @tarikstafford the chance to actually find it that high up is low, and quite expensive, so to avoid traversing the whole hierarchy unnecessarily I thought it was good to cap it :)

PR #435, opened 2021-12-21
Claim status detail screen

SatishBV:
## [APP-1135]

- Let's member see details of the claim status detail screen

## Checklist

- [x] Landscape verification
- [x] iPad verification

[APP-1135]: https://hedvig.atlassian.net/browse/APP-1135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Projects/Home/Sources/Claims/ClaimDetailView.swift
```diff
@@ -0,0 +1,107 @@
...
+            case .paid = claim.claimDetailData.outcome,
+            !claim.claimDetailData.payout.amount.isEmpty
+        {
+            return claim.claimDetailData.payout
+        }
+
+        return nil
+    }
+
+    private var contactChatView: some View {
```
sampettersson, line 36: I think this is a sign that this should be split into its own struct, right? 😄

Projects/Home/Sources/Claims/ClaimDetailHeader.swift
```diff
@@ -0,0 +1,137 @@
...
+    private var closedDate: String {
+        let dateFormatter = DateFormatter.withIso8601Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ")
+        guard let closed = closed,
+            let date = dateFormatter.date(from: closed)
+        else { return "-" }
+        return readableDateString(from: date)
+    }
+
+    /// Converts date into a readable friendly string
+    private func readableDateString(from date: Date) -> String {
```
sampettersson, line 44: I remember that we have some very similar code in hCore, maybe we can reuse that?

Projects/Home/Sources/Claims/ClaimDetailHeader.swift
```diff
@@ -0,0 +1,137 @@
...
+    private var submittedDate: String {
+        let dateFormatter = DateFormatter.withIso8601Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ")
+        guard let submitted = submitted,
+            let date = dateFormatter.date(from: submitted)
+        else { return "-" }
+        return readableDateString(from: date)
+    }
+
+    private var closedDate: String {
+        let dateFormatter = DateFormatter.withIso8601Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ")
```
sampettersson, line 36: Can we extend `String.localDateToDate` maybe?

Projects/Home/Sources/AudioTrack/TrackPlayer.swift
```diff
@@ -0,0 +1,128 @@
...
+
+    private func startPlaying() {
+        let session = AVAudioSession.sharedInstance()
+
+        do {
+            try session.setCategory(.playback)
+            try session.overrideOutputAudioPort(AVAudioSession.PortOverride.speaker)
+            try session.setActive(true)
+        } catch {
+            print("Playing over the device's speakers failed")
```
sampettersson, line 100: Should we show an error to the user?

sampettersson (changes requested):
Haven't checked on the phone yet but code looks super solid! Great job! Only thing is look through and try to split up the components a bit more, some of them are a bit chunky and hard to read :) also try to avoid multiple structs/classes in the same file.

Projects/Home/Sources/Claims/ClaimDetailHeader.swift
SatishBV: I needed to use RelativeDateTimeFormatter for this particular problem. I haven't found it to be used else where in the app. Do you know of any particular name by which I can find similar code?

If not, I can put this code in hCore

Projects/Home/Sources/AudioTrack/TrackPlayer.swift
SatishBV: I copied this code from Embark flow to initialise audio player. Chances of error here should be less.

But if an error  has to be shown, do you have any standard template code to show errors? Like should it be an alert or some other view?

Projects/App/Sources/Journeys/ClaimsJourney.swift
```diff
@@ -96,5 +96,11 @@ extension AppJourney {
         ) { action in
             DismissJourney()
         }
+        .addConfiguration { presenter in
+            presenter.viewController.navigationItem.largeTitleDisplayMode = .never
+            presenter.viewController.navigationItem.title = L10n.ClaimStatus.title
```
sampettersson, line 101: Title has a method `configureTitle` which you can use.

Projects/App/Sources/Journeys/ClaimsJourney.swift
```diff
@@ -96,5 +96,11 @@ extension AppJourney {
         ) { action in
             DismissJourney()
         }
+        .addConfiguration { presenter in
+            presenter.viewController.navigationItem.largeTitleDisplayMode = .never
```
sampettersson, line 100: Maybe good to break this out into a function and put it in hCore?

Projects/Home/Sources/Claims/ClaimDetailView.swift
```diff
@@ -79,19 +79,19 @@ public struct ClaimDetailView: View {
 
             Spacer()
         }
-        .navigationBarTitle(Text(L10n.ClaimStatus.title), displayMode: .inline)
         .onReceive(
-            store.stateSignal
-                .plain()
-                .distinct()
+            store
```
sampettersson: I think this logic can be solved neater with the `PresentableStoreLens`, check here for reference: https://github.com/HedvigInsurance/ugglan/blob/666c8bc60596482e28b1db53767dcb1ef77edfeb/Projects/Contracts/Sources/View/ContractTable.swift#L35

Projects/hGraphQL/Sources/Models/Claims.swift
```diff
@@ -200,3 +200,20 @@ public struct ClaimData {
         claims = cardData.claimsStatusCards.map { .init(cardData: $0) }
     }
 }
+
+extension Claim: Equatable {
+    public static func ==(lhs: Claim, rhs: Claim) -> Bool {
+        return lhs.id == rhs.id
```
sampettersson, line 206: Why not make the whole object equatable, why just check the id? Doing this usually leads to weird behaviours I've seen 😬

Projects/Home/Sources/Claims/ClaimDetailView.swift
```diff
@@ -0,0 +1,83 @@
...
+
+public struct ClaimDetailView: View {
+    @State var claim: Claim
+    var store: HomeStore
+
+    public init(
+        claim: Claim
+    ) {
+        self.claim = claim
+        let store: HomeStore = globalPresentableStoreContainer.get()
```
sampettersson, line 15: Why not use @PresentableStore to fetch this?

sampettersson (changes requested):

sampettersson (approved):

PR #452, opened 2022-01-26
Quote Cart

tarikstafford:
## [APP-1283]

- Add `CreateOnboardingQuoteCart` to Ugglan
- If the quoteCart feature flag is on we fetch create the quote cart and start the onboarding flow (need it before embark so we pick the correct embark flows)
- We remove the token if there is one when starting embark as well.
- Offer has been refactored to handle quote cart graphql queries and mutations.
- Signing flow is a bit different now we call `startCheckout` (not to be confused with checkout for NO or DK) and then we poll quote cart and wait for a checkout status of .signed and then create an access token and start `PostOnboarding` 

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

[APP-1283]: https://hedvig.atlassian.net/browse/APP-1283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Projects/Market/Sources/MarketStore.swift
```diff
@@ -2,28 +2,41 @@ import Flow
...
     public override func effects(
         _ getState: @escaping () -> MarketState,
         _ action: MarketAction
     ) -> FiniteSignal<MarketAction>? {
         switch action {
         case let .selectMarket(market):
             Localization.Locale.currentLocale = market.preferredLanguage
+            return self.client
+                .perform(
+                    mutation: GraphQL.CreateOnboardingQuoteCartMutation(input: .init(market: market.graphQL, locale: market.preferredLanguage.rawValue))
```
sampettersson, line 33: Doesn't this Query make more sense in `Offer`?

hugokallstrom: I'm going to start implementing this now on Android. Popping in here to bounce som ideas. 

We need to make this query before creating a quote in embark, this is the mutation specified in angel:
```
mutation CreateSwedishQuote(
    $swedishHomeQuoteId: ID!
    ...
)
```

Can we use `EmbarkStoryType` to know when we need to generate a quote cart id? We could then check `EmbarkStoryType.APP_ONBOARDING_QUOTE_CART`. We would then need to add more types for moving flow etc. But maybe thats ok? 

This could perhaps be done at the same time we are querying the story.

Thoughts?

tarikstafford: @sampettersson we need the `quoteCartID` in Embark unfortunately. I was talking with robin about adding this into embark as an `apiFragment`.

tarikstafford: @sampettersson I will move this query to embark. Do we want to tuck quote cart behind a feature flag? @hugokallstrom

PR #469, opened 2022-02-15
APP-911: In-app updates

axelbacklund:
## [APP-911]

- Alert screen for users that have old versions of the Hedvig app, based on an Unleash feature flag

## Checklist

- [x] Landscape verification
- [x] iPad verification

[APP-911]: https://hedvig.atlassian.net/browse/APP-911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Projects/App/Sources/Screens/UpdateAppScreen/UpdateAppView.swift
```diff
@@ -0,0 +1,46 @@
...
+import Presentation
+import SwiftUI
+import UIKit
+import hCore
+import hCoreUI
+import hGraphQL
+
+struct UpdateAppView: View {
+    var body: some View {
+        ZStack {
```
sampettersson, line 13: Maybe we should just use hForm and hRow here, so that this screen isn't too custom, what do you think?

sampettersson (approved):
Code looks great 👌 just a thought on using some hCoreUI primitives

Projects/App/Sources/Screens/UpdateAppScreen/UpdateAppView.swift
axelbacklund: Agree 👍

PR #498, opened 2022-04-06
Tracking events

sampettersson:
## [APP-1547]

> Some tracking events

[APP-1547]: https://hedvig.atlassian.net/browse/APP-1547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Projects/App/Sources/Screens/ClaimsScreen/Screens/HonestyPledge.swift
```diff
@@ -86,6 +86,7 @@ struct DidAcceptPledgeNotifier: View {
                 if value && !hasNotifiedStore {
                     hasNotifiedStore = true
                     store.send(.didAcceptHonestyPledge)
+                    hAnalyticsEvent.honorPledgeConfirmed().send()
```
SatishBV, line 89: I was wondering if we can start sending the `hAnalyticsEvents` from the Store inside `effects` instead. 

This keeps all the views from having to import `hAnalytics`. I also feel it is more appropriate to place it in `effects` because it is a side effect happening after an action .🤔

SatishBV (approved):

PR #528, opened 2022-05-17
Secure storage in Keychain

SatishBV:
## [APP-1648]

- Stores the auth-token in Keychain instead of disk

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

[APP-1648]: https://hedvig.atlassian.net/browse/APP-1648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sampettersson (commented):
Code looks great! Although we need some way to migrate over existing tokens right?

SatishBV:
> Code looks great! Although we need some way to migrate over existing tokens right?

You are right! Totally overlooked that! 
Should probably check back on disk for the first time when the keychain is empty and update keychain and remove from disk.
If disk is also empty, then it's a new user

Projects/hGraphQL/Sources/ApolloClient.swift
```diff
@@ -119,19 +121,36 @@ extension ApolloClient {
...
         Future { completion in
-            let tokenData = self.retreiveToken()
-
-            if tokenData == nil {
-                return self.createClientFromNewSession()
-                    .onResult { result in
-                        switch result {
-                        case let .success(result): completion(.success(result))
-                        case let .failure(error): completion(.failure(error))
+            let keychainToken = self.retreiveToken()
```
sampettersson, line 124:
```suggestion
            guard let keychainToken = self.retreiveToken() else { fallbackToDisk }
```

Maybe?

PR #531, opened 2022-05-18
APP-1531 Contracts display refactor

tarikstafford:
## [APP-1531]

- Removed `Contract Filter` instead we have two arrays of contracts in the `ContractStore` now.
- Created a new `TerminatedContracts` view to reduce noise.

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

[APP-1531]: https://hedvig.atlassian.net/browse/APP-1531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Projects/Contracts/Sources/Contracts.swift
```diff
@@ -85,6 +35,7 @@ extension Contracts: View {
             fetch()
         }
         .trackOnAppear(hAnalyticsEvent.screenView(screen: .insurances))
+        .navigationBarTitle(L10n.InsurancesTab.title)
```
sampettersson, line 38: This one usually leads to the title being set too late and it looks off, why did you move this away from the Journey?

Projects/Contracts/Sources/Contracts.swift
```diff
@@ -8,64 +8,14 @@ import hCore
...
-    let filter: ContractFilter
-
-    public init(
-        filter: ContractFilter
-    ) {
-        self.filter = filter
-    }
+    
+    @State
+    var navigationController: UINavigationController?
```
sampettersson, line 16: This is not used right?

sampettersson (changes requested):

Projects/Contracts/Sources/View/ContractTable.swift
```diff
@@ -9,82 +9,72 @@ import hCoreUI
...
-        case .active:
-            return state
-                .contractBundles
-                .flatMap { $0.contracts }
-        case .terminated:
-            return state.contracts.filter { contract in
-                contract.currentAgreement?.status == .terminated
-            }
-        case .none: return []
+        if contracts.isEmpty {
```
SatishBV, line 20: Maybe this? 
```suggestion
        self.contracts = contracts.isEmpty ? store.state.terminatedContracts : contracts
```

Projects/Contracts/Sources/View/ContractTable.swift
```diff
@@ -9,82 +9,72 @@ import hCoreUI
...
-                    getContractsToShow(for: state, filter: filter.nonemptyFilter(state: state))
-                }
-            ) { contracts in
-                ForEach(contracts, id: \.id) { contract in
-                    ContractRow(id: contract.id)
-                        .fixedSize(horizontal: false, vertical: true)
-                        .padding(.top, 15)
-                        .transition(.slide)
-                }
+            ForEach(contracts, id: \.id) { contract in
```
SatishBV, line 33: Does removing the `PresentableStoreLens` not have any affect when polling for contracts?

Projects/Contracts/Sources/View/ContractTable.swift
```diff
@@ -9,82 +9,72 @@ import hCoreUI
...
+        
+        hSection {
+            PresentableStoreLens(
+                ContractStore.self,
+                getter: { state in
+                    return state.terminatedContracs
+                }
+            ) { terminatedContracts in
+                if !terminatedContracts.isEmpty {
+                    hSection(header: hText(L10n.InsurancesTab.moreTitle)) {
```
SatishBV, line 55: Maybe move this view to a new property and avoid this indent gore? Not super important but felt it would be much more clearer to make sense of what is going on here

SatishBV (commented):
Super nice this is 🚀

Projects/Contracts/Sources/View/ContractTable.swift
tarikstafford: Not that I am aware of. The polling should update the `state` which is received in the `onReceive` block, which then updates the local `state` value.

Projects/Contracts/Sources/View/ContractTable.swift
tarikstafford: Yeah good call!

Projects/Contracts/Sources/Contracts.swift
tarikstafford: Can move it back, originally I was going to reuse the `ContractsTable` for `Terminated` as well but now I created a new view. So I can just move it back 👯

SatishBV (approved):

Projects/hCore/Sources/ScrollView+Inspect.swift
```diff
@@ -65,6 +65,83 @@ struct ViewIntrospector<ViewType: UIView>: UIViewRepresentable {
     }
 }
 
+class IntrospectionViewController: UIViewController {
```
sampettersson, line 68: Remove if you didn't end up using this maybe?

sampettersson (approved):

Projects/hCore/Sources/ScrollView+Inspect.swift
tarikstafford: 🆗

sampettersson (approved):

PR #564, opened 2022-06-21
Remove the check which doesn’t let icons update on changing traits

SatishBV:
## [APP-1476]

- The `RemoteVectorIcon` doesn't update on changing from light to dark and vice-versa. The check for imageView's size returns and doesn't let the icon render again.

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

[APP-1476]: https://hedvig.atlassian.net/browse/APP-1476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Projects/hCoreUI/Sources/RemoteVectorIcon.swift
+++ b/Projects/hCoreUI/Sources/RemoteVectorIcon.swift
@@ -45,8 +45,6 @@ extension RemoteVectorIcon: Viewable {
         func renderPdfDocument(pdfDocument: CGPDFDocument) {
             let imageViewSize = imageView.frame.size
 
-            if let image = imageView.image { if image.size == imageViewSize { return } }
-
             let page = pdfDocument.page(at: 1)!
             let rect = page.getBoxRect(CGPDFBox.mediaBox)
 


Projects/hCoreUI/Sources/RemoteVectorIcon.swift
```diff
@@ -45,8 +45,6 @@ extension RemoteVectorIcon: Viewable {
         func renderPdfDocument(pdfDocument: CGPDFDocument) {
             let imageViewSize = imageView.frame.size
 
-            if let image = imageView.image { if image.size == imageViewSize { return } }
```
sampettersson, line 48: Maybe we should check if traits actually changed? Not sure this will be super performant.

SatishBV: This `renderPdfDocument` was being called when traits/layout is changed. I think it should not affect the performance too much in that case 🤔

sampettersson: layout happens quite often actually 😓

sampettersson: We actually do a check here: https://github.com/HedvigInsurance/ugglan/blob/0961647812451371a404266b98772a7ca87b5df3/Projects/hCoreUI/Sources/RemoteVectorIcon.swift#L86

So I guess its fine to remove that line :)

PR #595, opened 2022-07-04
Validate auth token on app startup

SatishBV:
## [APP-1756]

- Validate auth token and logout the member if the token is found to be invalid

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

[APP-1756]: https://hedvig.atlassian.net/browse/APP-1756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Projects/App/Sources/UgglanStore.swift
+++ b/Projects/App/Sources/UgglanStore.swift
@@ -5,6 +5,7 @@ import Offer
 import Presentation
 import UIKit
 import hCore
+import hCoreUI
 import hGraphQL
 
 public struct UgglanState: StateProtocol {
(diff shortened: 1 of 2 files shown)


Projects/App/Sources/UgglanStore.swift
```diff
@@ -67,6 +69,28 @@ public final class UgglanStore: StateStore<UgglanState, UgglanAction> {
             return performTokenExchange(with: exchangeToken)
         case let .exchangePaymentToken(token):
             return performTokenExchange(with: token)
+        case .validateAuthToken:
+            return client.fetch(query: GraphQL.MemberIdQuery())
+                .valueThenEndSignal
+                .atError(on: .main) { error in
+                    print(error.localizedDescription)
```
sampettersson, line 76: remove this, but we should probably log here so we can see in Datadog if this happens too often.

sampettersson (changes requested):

sampettersson (approved):

PR #655, opened 2023-02-23
Customer io pushes

sampettersson:
Register push tokens in Customer IO instead of Firebase


sampettersson:
Add a listener for has token

sladan-hedvig (approved):
I tried to send myself notification after registering thought chat but never received it 🐹 
Others we are sending to BE when we have it 👍

sampettersson:
@sladan-hedvig just to make sure:

1. Push permissions was accepted
2. You where using a Release build on a physical device (TestFlight works fine, but not a local debug build that goes to the APNS sandbox)
3. The app wasn't in the foreground

If it still fails after all of these are fulfilled please let me know.

sladan-hedvig:
@sampettersson  I used debug version. I guess thats the issue ^^

sladan-hedvig:
Its working! Awesome!

PR #682, opened 2023-04-05
Hide coinsured for non applicable contracts

sampettersson:

### DIFF
--- a/Projects/hGraphQL/Sources/Models/ContractModels.swift
+++ b/Projects/hGraphQL/Sources/Models/ContractModels.swift
@@ -40,6 +40,7 @@ extension String {
 public struct Contract: Codable, Hashable, Equatable {
     public init(
         id: String,
+        typeOfContract: TypeOfContract,
         upcomingAgreementsTable: DetailAgreementsTable,
         currentAgreementsTable: DetailAgreementsTable?,
         gradientOption: Contract.GradientOption?,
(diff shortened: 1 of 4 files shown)


Projects/hGraphQL/Sources/Models/ContractModels.swift
```diff
@@ -175,6 +180,150 @@ public struct Contract: Codable, Hashable, Equatable {
...
+            return true
+        case .dkTravel:
+            return true
+        case .dkTravelStudent:
+            return true
+        }
+    }
+}
+
+public struct TerminationStartFlow: Codable, Hashable, Equatable {
```
sampettersson, line 313: Wait why is this in my diff 🧐

Projects/hGraphQL/Sources/Models/ContractModels.swift
```diff
@@ -135,6 +139,7 @@ public struct Contract: Codable, Hashable, Equatable {
         contract: GiraffeGraphQL.ContractsQuery.Data.Contract
     ) {
         id = contract.id
+        typeOfContract = TypeOfContract(rawValue: contract.typeOfContract.rawValue)!
```
sladan-hedvig, line 142: if you make typeOfContract optional we don't need ! 
or 
if it stays required you need to place " ?? .unknown" instead !

sladan-hedvig (changes requested):

Projects/hGraphQL/Sources/Models/ContractModels.swift
sampettersson: Yeah, crossed my mind actually adding .unknown sounds good.

PR #687, opened 2023-04-05
Added notification permission screen to claims flow

sladan-hedvig:
## [CLX-382]

- Reentered notification screen for claims flow

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

[CLX-382]: https://hedvig.atlassian.net/browse/CLX-382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Projects/App/Sources/Journeys/FeaturesLoader.swift
```diff
@@ -83,3 +83,23 @@ struct NotificationLoader: Presentable {
...
+            switch status {
+            case .notDetermined:
+                break
+            default:
+                let store: ClaimsStore = globalPresentableStoreContainer.get()
+                store.send(.startClaim(from: originId))
+            }
+            semaphore.signal()
+        }
+        semaphore.wait()
```
sampettersson, line 102: This would block the thread, riight? 🤔 Check `NotificationLoader`, it might help you do this by adding a loader screen in between, or we should track this in some store to make sure we don't need to synchronously await an async transaction.

Projects/App/Sources/Journeys/ClaimsJourney.swift
```diff
@@ -86,15 +76,29 @@ extension AppJourney {
...
+                            }
                     }
+                }
             } else {
                 ClaimJourneys.getScreenForAction(for: action, withHidesBack: true)
             }
         }
+        .onAction(ClaimsStore.self) { action, nav in
+            if case .startClaim = action {
+                nav.viewController.navigationController?.popToViewController(nav.viewController, animated: true)
```
sampettersson, line 98: This seems odd to me, why did you have to do this?

sampettersson (changes requested):

Projects/App/Sources/Journeys/FeaturesLoader.swift
sladan-hedvig: Yeah, it will block, but it is like 0.1 seconds and caused me a lot of issues. Maybe if you have some time tomorrow, we can go thought it together.

Projects/App/Sources/Journeys/ClaimsJourney.swift
sladan-hedvig: Removed!

sampettersson (approved):

PR #699, opened 2023-04-12
Feature/CLX-521 refactor money fragment

julia-andersson:
## [CLX-521]

- Refactored code for formatting double

## Checklist

- [X] Dark/light mode verification
- [X] Landscape verification
- [X] iPad verification
- [X] iPod/iPhone 12 mini/iPhone SE verification

[CLX-521]: https://hedvig.atlassian.net/browse/CLX-521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
+++ b/Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
@@ -42,28 +42,11 @@ struct ClaimFlowMoneyModel: Codable, Equatable {
 
     func getAmountWithCurrency() -> String {
         let formatter = NumberFormatter()
-        formatter.groupingSeparator = ""
         formatter.numberStyle = .currency
         formatter.minimumFractionDigits = 0
         formatter.maximumFractionDigits = 2
         formatter.currencyCode = self.currencyCode
         formatter.currencySymbol = self.currencyCode
         return formatter.string(for: amount) ?? ""
-    }
-
-    func formatAndPrintDouble(value: Double) -> String {
-        let formatter = NumberFormatter()
-        formatter.groupingSeparator = " "
-        formatter.numberStyle = .decimal
-
-        if value.truncatingRemainder(dividingBy: 1) == 0 {
-            formatter.minimumFractionDigits = 0
-            formatter.maximumFractionDigits = 0
-        } else {
-            formatter.minimumFractionDigits = 1
-            formatter.maximumFractionDigits = 2
-            formatter.decimalSeparator = "."
-        }
-        return formatter.string(for: value) ?? ""
     }
 }
--- a/Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
+++ b/Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
@@ -50,4 +50,20 @@ struct ClaimFlowMoneyModel: Codable, Equatable {
         formatter.currencySymbol = self.currencyCode
         return formatter.string(for: amount) ?? ""
     }
+
+    func formatAndPrintDouble(value: Double) -> String {
+        let formatter = NumberFormatter()
+        formatter.groupingSeparator = " "
+        formatter.numberStyle = .decimal
+
+        if value.truncatingRemainder(dividingBy: 1) == 0 {
+            formatter.minimumFractionDigits = 0
+            formatter.maximumFractionDigits = 0
+        } else {
+            formatter.minimumFractionDigits = 1
+            formatter.maximumFractionDigits = 2
+            formatter.decimalSeparator = "."
+        }
+        return formatter.string(for: value) ?? ""
+    }
 }
--- a/Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
+++ b/Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
@@ -1,12 +1,13 @@
 import Foundation
+import hCore
 import hGraphQL
 
 public struct FlowClaimSingleItemCheckoutStepModel: FlowClaimStepModel {
     let id: String
-    let deductible: ClaimFlowMoneyModel
-    let depreciation: ClaimFlowMoneyModel
-    let payoutAmount: ClaimFlowMoneyModel
-    let price: ClaimFlowMoneyModel
+    let deductible: MonetaryAmount
+    let depreciation: MonetaryAmount
+    let payoutAmount: MonetaryAmount
+    let price: MonetaryAmount
     let payoutMethods: [AvailableCheckoutMethod]
     var selectedPayoutMethod: AvailableCheckoutMethod?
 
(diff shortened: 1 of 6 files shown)


Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
```diff
@@ -50,4 +50,20 @@ struct ClaimFlowMoneyModel: Codable, Equatable {
         formatter.currencySymbol = self.currencyCode
         return formatter.string(for: amount) ?? ""
     }
+
+    func formatAndPrintDouble(value: Double) -> String {
+        let formatter = NumberFormatter()
+        formatter.groupingSeparator = " "
+        formatter.numberStyle = .decimal
```
sladan-hedvig, line 57: Try switching it to .currency and set currencySymbol for formatter.
You don't need to send value since we can do any operation on amount property inside structure.
Then, this function will return something like 1.200 SEK

Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
```diff
@@ -50,4 +50,20 @@ struct ClaimFlowMoneyModel: Codable, Equatable {
...
+
+        if value.truncatingRemainder(dividingBy: 1) == 0 {
+            formatter.minimumFractionDigits = 0
+            formatter.maximumFractionDigits = 0
+        } else {
+            formatter.minimumFractionDigits = 1
+            formatter.maximumFractionDigits = 2
+            formatter.decimalSeparator = "."
```
sladan-hedvig, line 65: I don't think that we should define decimal seperator.
Also I don't think you need to check if  truncatingRemainder == 0. Just need to set minimumFractionDigits to 0 and maximum to 2

Projects/Claims/Sources/Views/submitClaim/Checkout/SubmitClaimCheckoutNoRepairScreen.swift
```diff
@@ -88,50 +87,15 @@ public struct SubmitClaimCheckoutNoRepairScreen: View {
...
-                    .foregroundColor(hLabelColor.secondary)
-                } else {
-                    hText(
-                        prefix + formatDoubleWithDecimal(value: model?.amount ?? 0) + " "
-                            + String(model?.currencyCode ?? "")
-                    )
-                    .foregroundColor(hLabelColor.secondary)
-                }
+                hText(
+                    prefix + (model?.formatAndPrintDouble(value: model?.amount ?? 0) ?? "") + " "
```
sladan-hedvig, line 91: Then after changes we should have hText(prefix + (model?.formatAndPrintDouble() ?? ""))

Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
```diff
@@ -50,4 +50,20 @@ struct ClaimFlowMoneyModel: Codable, Equatable {
         formatter.currencySymbol = self.currencyCode
         return formatter.string(for: amount) ?? ""
     }
+
+    func formatAndPrintDouble(value: Double) -> String {
```
sladan-hedvig, line 54: maybe we should rename this function to something like
func getValueWIthCurrency() -> String

sladan-hedvig (approved):

Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
```diff
@@ -42,7 +42,6 @@ struct ClaimFlowMoneyModel: Codable, Equatable {
 
     func getAmountWithCurrency() -> String {
```
sampettersson, line 50: This could be a dynamic var as its not taking any arguments :)

Projects/Claims/Sources/Views/submitClaim/Checkout/SubmitClaimCheckoutNoRepairScreen.swift
```diff
@@ -88,50 +87,14 @@ public struct SubmitClaimCheckoutNoRepairScreen: View {
...
-                    .foregroundColor(hLabelColor.secondary)
-                } else {
-                    hText(
-                        prefix + formatDoubleWithDecimal(value: model?.amount ?? 0) + " "
-                            + String(model?.currencyCode ?? "")
-                    )
-                    .foregroundColor(hLabelColor.secondary)
-                }
+                hText(
+                    prefix + (model?.getAmountWithCurrency() ?? "")
```
sampettersson, line 81: Instead of using prefix here, couldn't we just negate the amount? And pass a negative amount to the formatter then it should add the `-1000kr` automatically 😄

Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
```diff
@@ -42,7 +42,6 @@ struct ClaimFlowMoneyModel: Codable, Equatable {
...
         let formatter = NumberFormatter()
-        formatter.groupingSeparator = ""
         formatter.numberStyle = .currency
         formatter.minimumFractionDigits = 0
```
sampettersson, line 53: Don't forget to pass locale into this, else it will use the device one, which might not be the one we want

I think its something like this:

```suggestion
        formatter.locale = Localization.Locale.currentLocale.foundation
```

sampettersson (changes requested):

Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
sladan-hedvig: I think thats the way to go. We should adjust to phone settings, by my opinion.
So currency decimal, group separator should be from system.
We can add it if required. 😸

Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
sladan-hedvig: 👍

Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
julia-andersson: Shouldn't the locale be based on the account and not the device? Like Sam said

sampettersson: Some people have en_US on their phone but the app in `sv_SE` which would entail something like `10,000 SEK` which becomes rather odd for the user, so in all other places we've just passed in the locale from Localization.Locale 🙈

sladan-hedvig: Oki  😄  thats fine if required.

Projects/Claims/Sources/Views/submitClaim/Checkout/SubmitClaimCheckoutNoRepairScreen.swift
julia-andersson: @sampettersson The thing is that we then have to "modify" the values that we get from the backend to be negative which can cause problems in the future if we need to use these values somewhere else  😅

Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
sampettersson: Hold on, I just realised, we already have a type `MonetaryAmount`, which currently can just be constructed from `MonetaryAmountFragment` could we maybe just make a `MoneyFragment` construct a `MonetaryAmount`, then we are 100% sure we use the same logic everywhere for this, and don't have to duplicate formatters 😄

Projects/Claims/Sources/Views/submitClaim/Checkout/SubmitClaimCheckoutNoRepairScreen.swift
sampettersson: I wouldn't store them as negative, that I agree with can cause problems, but if we know that something always means negative from a user POV, like a deductible or a discount, then I think it makes sense to do that conversion in the view layer, totally agree to not store it like a negative amount in the Store.

Projects/Claims/Sources/Models/FlowClaimSingleItemCheckoutStepModel.swift
julia-andersson: Used that now! :)

sampettersson (approved):
Cleeeaan ⭐  great job!

PR #735, opened 2023-04-25
Fix views not opening correctly after clicking push notification

sampettersson:
Fixes some issues i found while testing pushes

### DIFF
--- a/Projects/App/Sources/AppDelegate+Notifications.swift
+++ b/Projects/App/Sources/AppDelegate+Notifications.swift
@@ -42,6 +42,83 @@ extension AppDelegate {
 }
 
 extension AppDelegate: UNUserNotificationCenterDelegate {
+    fileprivate func performPostLoggedIn(work: @escaping () -> Void) {
+        bag += ApplicationContext.shared.$isLoggedIn.atOnce().filter { $0 }
+            .onFirstValue { _ in
+                work()
+            }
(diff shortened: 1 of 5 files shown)


Projects/App/Sources/AppDelegate+Notifications.swift
```diff
@@ -42,6 +42,83 @@ extension AppDelegate {
...
+            }
+    }
+
+    fileprivate func performPushAction(notificationType: String, userInfo: [AnyHashable: Any]) {
+        hAnalyticsEvent.notificationOpened(type: notificationType).send()
+
+        if notificationType == "NEW_MESSAGE" {
+            performPostLoggedIn {
+                let store: UgglanStore = globalPresentableStoreContainer.get()
+                store.send(.openChat)
```
sladan-hedvig, line 58: We need to add check if the chat screen is already opened, so we don't open it twice.
It will not open over the current chat, but will show again after we dismiss chat.

sladan-hedvig (changes requested):

Projects/App/Sources/AppDelegate+Notifications.swift
sladan-hedvig: Awesome! It works fine!

sladan-hedvig (approved):

PR #745, opened 2023-05-05
Add Squircle

sampettersson:
This adds a new shape, which replaces `.defaultCornerRadius` with a Square with kind of round corners, but also not really, they are pretty cool 😄 this is in accordance with the new design system.

Simulator Screenshot - iPhone 14 Pro - 2023-05-05 at 11 45 03


sladan-hedvig (approved):
Cool 😎

PR #770, opened 2023-05-31
added blur and scoll option to hForm

julia-andersson:
## [APP-XXX]

- Change 1
- Change 2

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/hCoreUI/Sources/hForm/hForm.swift
+++ b/Projects/hCoreUI/Sources/hForm/hForm.swift
@@ -146,18 +146,27 @@ public struct hForm<Content: View>: View {
                     BackgroundView().edgesIgnoringSafeArea(.all)
                 }
             }
-            if hDisableScroll {
-                displayContent
-                    .frame(maxHeight: .infinity, alignment: .top)
-            } else {
-                ScrollView {
-                    displayContent
-                }
-                .modifier(ForceScrollViewIndicatorInset(insetBottom: bottomAttachedViewHeight))
-                .findScrollView { scrollView in
-                    if #available(iOS 15, *) {
-                        scrollView.viewController?.setContentScrollView(scrollView)
+            ScrollView {
+                VStack {
+                    if let hFormTitle, hUseNewStyle {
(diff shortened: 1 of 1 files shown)


Projects/hCoreUI/Sources/hForm/hForm.swift
```diff
@@ -110,26 +140,24 @@ public struct hForm<Content: View>: View {
...
-                .frame(maxWidth: .infinity)
-                .tint(hForm<Content>.returnTintColor(useNewStyle: hUseNewStyle))
-                Color.clear
-                    .frame(height: bottomAttachedViewHeight)
             }
-            .modifier(ForceScrollViewIndicatorInset(insetBottom: bottomAttachedViewHeight))
-            .findScrollView { scrollView in
-                if #available(iOS 15, *) {
-                    scrollView.viewController?.setContentScrollView(scrollView)
+            if hDisableScroll {
```
sladan-hedvig, line 149: Instaed of having different content depending on hDisableScroll (since we are using scrollView when we are using .detent style) i suggest to do it like this:
            .findScrollView { scrollView in
                if #available(iOS 15, *) {
                    scrollView.viewController?.setContentScrollView(scrollView)
                }
                if hDisableScroll {
                    scrollView.addGestureRecognizer(UIPanGestureRecognizer())
                }
            }

Projects/hCoreUI/Sources/hForm/hForm.swift
```diff
@@ -174,3 +218,20 @@ extension View {
...
+
+extension EnvironmentValues {
+    public var hUseBlur: Bool {
+        get { self[EnvironmentHUseBlur.self] }
+        set { self[EnvironmentHUseBlur.self] = newValue }
+    }
+}
+
+extension View {
+    public var hUseBlur: some View {
```
sladan-hedvig, line 234: We should be able to use hUseBlur on every type of style we want to use. 
We need to implement it in the PresenationOptions in Presentation Library

PR #842, opened 2023-06-26
Updated color and fixed buttons that did not look as expected

julia-andersson:
## [APP-XXX]

- Added different options for LoadingButtonWithContent to match date picker need
- Updated color from lavender to black

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/hCoreUI/Sources/Styling/DefaultStyling.swift
+++ b/Projects/hCoreUI/Sources/Styling/DefaultStyling.swift
@@ -198,10 +198,10 @@ extension DefaultStyling {
             for: .highlighted
         )
 
-        UIDatePicker.appearance().tintColor = .brand(.link)
+        UIDatePicker.appearance().tintColor = .brandNew(.primaryText())
 
         UIImageView.appearance().tintColor = .brand(.primaryTintColor)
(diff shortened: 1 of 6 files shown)


Projects/hCoreUI/Sources/Styling/DefaultStyling.swift
```diff
@@ -198,10 +198,10 @@ extension DefaultStyling {
             for: .highlighted
         )
 
-        UIDatePicker.appearance().tintColor = .brand(.link)
+        UIDatePicker.appearance().tintColor = .brandNew(.primaryText())
```
julia-andersson, line 201: @sladan-hedvig not sure if we should just change the date picker color for both old and new design like this or somehow check if new design is enabled?

sladan-hedvig: Returned to previous value. Updated tint color in the place

sladan-hedvig (approved):

Projects/hCoreUI/Sources/Styling/DefaultStyling.swift
julia-andersson: @sladan-hedvig but date picker is still black for old design. should it be that?

PR #847, opened 2023-06-27
updated business model to not contain cashback

julia-andersson:
## [APP-XXX]

- Updated business model to not contain cashback

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/App/Sources/Screens/ProfileScreen/ProfileState.swift
+++ b/Projects/App/Sources/Screens/ProfileScreen/ProfileState.swift
@@ -28,7 +28,7 @@ public enum ProfileAction: ActionProtocol {
     case openFreeTextChat
     case openAppInformation
     case openAppSettings
-    case setProfileState(name: String, charity: String, monthlyNet: Int)
+    case setProfileState(name: String, monthlyNet: Int)
     case setEurobonusNumber(partnerData: PartnerData?)
     case fetchProfileStateCompleted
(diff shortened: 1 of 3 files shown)


Projects/App/Sources/Screens/ProfileScreen/ProfileState.swift
```diff
@@ -112,9 +111,8 @@ public final class ProfileStore: StateStore<ProfileState, ProfileAction> {
     public override func reduce(_ state: ProfileState, _ action: ProfileAction) -> ProfileState {
         var newState = state
         switch action {
-        case .setProfileState(let name, let charity, let monthlyNet):
+        case .setProfileState(let name, let monthlyNet):
             newState.memberFullName = name
-            newState.memberCharityName = charity
```
sladan-hedvig, line 117: can we also remove memberCharityName from state since we are not settings it ( old users will still have it set and will not change overtime) ?

sladan-hedvig (changes requested):

Projects/App/Sources/Screens/ProfileScreen/ProfileState.swift
julia-andersson: Yes I think we could do that. Good catch 👍

sladan-hedvig (approved):

PR #976, opened 2023-09-13
Rework button architecture

julia-andersson:
## [APP-XXX]

- Rework button structure to remove a lot of code

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/hCoreUI/Sources/hButton/hButton.swift
+++ b/Projects/hCoreUI/Sources/hButton/hButton.swift
@@ -13,10 +13,8 @@ struct LargeButtonModifier: ViewModifier {
 struct MediumButtonModifier: ViewModifier {
     func body(content: Content) -> some View {
         content
-            .padding(.horizontal, 16)
-            .padding(.vertical, 8)
-            .frame(minHeight: 32)
-
+            .frame(minHeight: 40)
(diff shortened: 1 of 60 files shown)


Projects/hCoreUI/Sources/hButton/hButton.swift
```diff
@@ -13,10 +13,9 @@ struct LargeButtonModifier: ViewModifier {
 struct MediumButtonModifier: ViewModifier {
     func body(content: Content) -> some View {
         content
+            .frame(minHeight: 40)
+            .frame(maxWidth: .infinity)
             .padding(.horizontal, 16)
-            .padding(.vertical, 8)
```
sladan-hedvig, line 17: I think we should have this. if for example content inside of a button has a height of 50, we will have no top and bottom spacing.

sladan-hedvig (changes requested):

Projects/hCoreUI/Sources/hButton/hButton.swift
julia-andersson: right. will add back 👍

sladan-hedvig (approved):

PR #991, opened 2023-09-22
GEN-1176: Update UI for moving flow

sladan-hedvig:
## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/MoveFlow/Sources/View/MovingFlowHousingType.swift
```diff
@@ -3,54 +3,74 @@ import SwiftUI
...
 
     public var body: some View {
-
         LoadingViewWithContent(MoveFlowStore.self, [.fetchMoveIntent], [.getMoveIntent]) {
-            hForm {
-                VStack {
-                    ForEach(housingTypes, id: \.self) { type in
-                        setCheckBoxComponent(text: type)
+            hForm {}
+                .hFormTitle(.standard, .title3, L10n.changeAddressSelectHousingTypeTitle)
```
julia-andersson, line 12: The text font doesn't look the same as in the figma design. Should be standard like it is on the other places

Projects/MoveFlow/Sources/View/MovingFlowApartmentView.swift
```diff
@@ -0,0 +1,216 @@
...
+        hFloatingTextField(
+            masking: Masking(type: .address),
+            value: $vm.address,
+            equals: $vm.type,
+            focusValue: .address,
+            error: $vm.addressError
+        )
+    }
+
+    func postalAndSquareField() -> some View {
```
julia-andersson, line 49: in the design we have "m2" (square meters) in the right placeholder which is always there. Can we add this?

Projects/MoveFlow/Sources/View/MovingFlowApartmentView.swift
```diff
@@ -0,0 +1,216 @@
...
+    @StateObject var vm = MovingFlowViewModel()
+
+    var body: some View {
+        hForm {}
+            .hFormTitle(.standard, .title1, L10n.changeAddressEnterNewAddressTitle)
+            .sectionContainerStyle(.transparent)
+            .hFormAttachToBottom {
+                hSection {
+                    VStack(spacing: 16) {
+                        VStack(spacing: 4) {
```
julia-andersson, line 17: When clicking on a field I have no way to dismiss the keyboard which covers the "send in" button :/

Projects/MoveFlow/Sources/View/MovingFlowProcessingView.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private var successView: some View {
+        ZStack(alignment: .bottom) {
+            BackgroundView().ignoresSafeArea()
+            VStack {
+                Spacer()
+                Spacer()
+                VStack(spacing: 16) {
+                    Image(uiImage: hCoreUIAssets.tick.image)
```
julia-andersson, line 39: Looks a bit too small. Make bigger according to design :)

julia-andersson (changes requested):

Projects/MoveFlow/Sources/View/MovingFlowApartmentView.swift
sladan-hedvig: Should be fine now 👍

PR #1027, opened 2023-10-12
Maker days/input text field

sladan-hedvig:
- Maker days input text field added

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Claims/Sources/ClaimStore.swift
```diff
@@ -30,13 +30,11 @@ public final class ClaimsStore: StateStore<ClaimsState, ClaimsAction> {
                         callback(.value(ClaimsAction.setClaims(claims: claimData)))
                     }
                     .onError { error in
-                        if ApplicationContext.shared.isDemoMode {
-                            callback(.value(.setLoadingState(action: action, state: nil)))
-                        } else {
-                            callback(
-                                .value(.setLoadingState(action: action, state: .error(error: L10n.General.errorBody)))
+                        callback(
```
julia-andersson, line 33: Why did we check for demo mode here?

Projects/Claims/Sources/SubmitClaimStore.swift
```diff
@@ -112,6 +112,15 @@ public final class SubmitClaimStore: LoadingStateStore<SubmitClaimsState, Submit
 
                 return disposeBag
             }
+        case let .submitTextInput(text):
```
julia-andersson, line 115: Do we really need a new action (.submitTextInput) here or can we reuse .submitAudioRecording in any way since we are calling the same mutation? Maybe send in a boolean argument as writtenText?

julia-andersson (changes requested):

Projects/Claims/Sources/ClaimStore.swift
sladan-hedvig: Because we dont want to show error if this fail for demo mode and it will since we dont have proper token

Projects/Claims/Sources/SubmitClaimStore.swift
sladan-hedvig: Yeah, maybe we can do it with enum. Give me a sec

Projects/Claims/Sources/ClaimStore.swift
julia-andersson: ah I see

julia-andersson (approved):

PR #1076, opened 2023-10-26
GEN-1030: Feature/bazooka/payments

sladan-hedvig:
- Updated schema
- Removed adyen

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Payment/Sources/Models/PaymentStatusData.swift
```diff
@@ -0,0 +1,54 @@
+import Foundation
+import hGraphQL
+
+public struct PaymentStatusData: Codable, Equatable {
+    public var status: PayinMethodStatus = .active
+    public var failedCharges: Int?  //TODO: FIX this
```
julia-andersson, line 6: What to fix? Is it something that should be fixed from backend?

sladan-hedvig: We should handle it somehow but will release as it is now. WIl update with new payments views

PR #1092, opened 2023-11-13
Feature/edit coinsured/GEN-451 deep link

julia-andersson:
## [GEN-451]

- Deep link for edit co-insured

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/App/Sources/AppDelegate+Deeplinks.swift
+++ b/Projects/App/Sources/AppDelegate+Deeplinks.swift
@@ -1,3 +1,4 @@
+import Contracts
 import CoreDependencies
 import Flow
 import Foundation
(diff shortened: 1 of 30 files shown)


Projects/App/Sources/Screens/ChatScreen/Models/Message.swift
```diff
@@ -249,6 +251,12 @@ func hash(into hasher: inout Hasher) { hasher.combine(globalId) }
                 type = .gif(url: URL(string: text.text))
             } else if text.text.isCrossSell {
                 type = .crossSell(url: URL(string: text.text))
+            } else if text.text.isDeepLink {
+                if let url = URL(string: text.text) {
```
sladan-hedvig, line 255: make sure (since we don't support other deep links handling) to set type to deepLink only for contract deep link.
In other cases is still should be text.

Projects/App/Sources/Screens/ChatScreen/Reusables/Message+Reusable.swift
```diff
@@ -95,6 +97,27 @@ extension Message: Reusable {
...
+            let imageHeight: CGFloat = {
+                let imageToShow = hCoreUIAssets.pillowHome.image
+                let ratio = imageToShow.size.width / imageToShow.size.height
+                return (227.77) / ratio
+            }()
+            if isRelatedToPreviousMessage {
+                return imageHeight + smallerMarginTop + extraHeightForTimeStampLabel
+            }
+
+            return imageHeight + largerMarginTop + extraHeightForTimeStampLabel
```
sladan-hedvig, line 119: height is not calculated right. We aren't taking title height in consideration and we are probably wrongly calculating imageHeight  - so it looks good now - and missed fixed size for spacing.

sladan-hedvig (approved):

PR #1097, opened 2023-11-17
GEN-1470: octopus payments history implementation

sladan-hedvig:
- octopus payment history implementation

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/hCore/Sources/Date+LocalDateString.swift
```diff
@@ -25,6 +25,10 @@ extension Date {
         return DateFormatters.displayddMMM.string(from: self)
     }
 
+    public var dateYYYYFormat: String? {
```
julia-andersson, line 28: @sladan-hedvig Why are some of these date formatters returning optional and some not? I think we should be consistent if there are not any special reason for this

julia-andersson (approved):

Projects/hCore/Sources/Date+LocalDateString.swift
sladan-hedvig: It shouldn't be optional :S lets fix that in another branch
https://media.tenor.com/HEMQmYHww6MAAAAd/borat-borat-very-nice.gif

PR #1126, opened 2023-11-30
fix for new schema

julia-andersson:
## [APP-XXX]

- Fix for co-insured being on contract instead of agreement

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Contracts/Sources/Models/ContractModels.swift
+++ b/Projects/Contracts/Sources/Models/ContractModels.swift
@@ -61,7 +61,8 @@ public struct Contract: Codable, Hashable, Equatable {
         firstName: String,
         lastName: String,
         ssn: String?,
-        typeOfContract: TypeOfContract
+        typeOfContract: TypeOfContract,
+        coInsured: [CoInsuredModel]
     ) {
         self.id = id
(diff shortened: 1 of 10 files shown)


Projects/Contracts/Sources/Models/ContractModels.swift
```diff
@@ -513,8 +510,8 @@ extension InsuredPeopleConfig {
     ) {
         let store: ContractStore = globalPresentableStoreContainer.get()
         self.init(
-            currentAgreementCoInsured: contract.currentAgreement?.coInsured ?? [],
-            upcomingAgreementCoInsured: contract.upcomingChangedAgreement?.coInsured,
+            currentAgreementCoInsured: contract.coInsured,
+            upcomingAgreementCoInsured: contract.coInsured,
```
sladan-hedvig, line 514: since currentAgreementCoInsured && upcomingAgreementCoInsured are same, we can have one property here like:
currentCoinsured

This will probably affect logic on some parts, but eventually make it easier to understand and read

sladan-hedvig (changes requested):

sladan-hedvig (approved):

PR #1138, opened 2023-12-06
GEN-1585: add hRadioOptionField

julia-andersson:
## [GEN-1585]

- Added new component hRadioOptionField
- [Figma](https://www.figma.com/file/WlwFCAWywXNcXVgPfYVmWu/Final-Design?node-id=7436%3A11028&mode=dev)

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/hCoreUI/Sources/hForm/RadioFields/hRadioOptionField.swift
```diff
@@ -0,0 +1,91 @@
+import SwiftUI
+import hCore
+
+public struct hRadioOptionField: View {
+    private var placeholder: String
+    private var value: String
```
sladan-hedvig, line 6: Can we change this to @Binding var value: String?

Projects/hCoreUI/Sources/hForm/RadioFields/hRadioOptionField.swift
```diff
@@ -0,0 +1,91 @@
...
+    @State private var animate = false
+    private var useAnimation: Bool
+    @Binding var error: String?
+
+    @State var selected: String = ""
```
sladan-hedvig, line 11: Instead of using this selected, use value.

Projects/hCoreUI/Sources/hForm/RadioFields/hRadioOptionField.swift
```diff
@@ -0,0 +1,91 @@
...
+            hFieldLabel(
+                placeholder: placeholder,
+                animate: $animate,
+                error: $error,
+                shouldMoveLabel: shouldMoveLabel
+            )
+            .padding(.bottom, 13)
+
+            HStack(spacing: 16) {
+                getCheckBox(texts: [L10n.General.yes, L10n.General.no])
```
sladan-hedvig, line 43: Try to pass the list of options when init `hRadioOptionField`. 
Maybe have defaults to [L10n.General.yes, L10n.General.no]

Projects/hCoreUI/Sources/hForm/RadioFields/hRadioOptionField.swift
```diff
@@ -0,0 +1,91 @@
...
+                Spacer()
+            }
+        }
+        .padding(.vertical, 16)
+        .addFieldBackground(animate: $animate, error: $error)
+        .padding(.horizontal, 16)
```
sladan-hedvig, line 49: This component shouldn't take care of horizontal spacing. Components that wrap this should be in charge of this:
.padding(.horizontal, 16)

In our case hSection will cover this

Projects/hCoreUI/Sources/CheckboxPickerScreen.swift
```diff
@@ -66,7 +69,14 @@ public struct CheckboxPickerScreen<T>: View where T: Equatable & Hashable {
         VStack(spacing: 4) {
             ForEach(items, id: \.object) { item in
                 hSection {
-                    getCell(item: item)
+                    HStack {
+                        if items.count < 3 {
+                            getCell(item: item)
+                                .frame(height: 72)
```
sladan-hedvig, line 75: Instead of having this set to 72, try to pass hFieldSize to function getCell.
hFieldSize is an enum

Based on hFieldSize property inside getCell function try to adjust paddings. 
Why?
Since we are having dynamic font, we could end up having not big enough field to fit text and it might look odd 👽

sladan-hedvig (approved):

PR #1186, opened 2024-01-19
Feature/bazooka/chat

sladan-hedvig:
- Reworked chat

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/App/Sources/Journeys/LoggedInJourney.swift
```diff
@@ -65,6 +67,15 @@ extension AppJourney {
             .configureSubmitClaimsNavigation
             .configurePaymentNavigation
             .configureContractNavigation
+            .configureChatNavigation
+        //            .onPresent {
+        //                DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
```
julia-andersson, line 72: remove comment

Projects/Chat/Sources/Views/ImagesView.swift
```diff
@@ -0,0 +1,294 @@
...
+                                    )
+                                    inCont.resume(returning: file)
+                                })
+                        }
+                case .image:
+                    guard let uti = contentInput?.uniformTypeIdentifier else {
+                        inCont.resume(throwing: GenerateFileUploadError.failedToGenerateMimeType)
+                        return
+                    }
+                    //
```
julia-andersson, line 228: remove

Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -0,0 +1,227 @@
...
+                await self?.fetch()
+            }
+        }
+        Task { [weak self] in
+            await self?.fetch()
+        }
+        let fileUploadManager = FileUploadManager()
+        fileUploadManager.resetuploadFilesPath()
+        DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
+            AskForRating().ask()
```
julia-andersson, line 42: Is this one ever triggered? Does not seem like it. Maybe we can either we can remove this or make it work?

julia-andersson (changes requested):

Projects/Chat/Sources/Views/ChatScreenViewModel.swift
sladan-hedvig: It is working, on third time if you check implementation
```
 func ask() {
        guard !UserDefaults.standard.bool(forKey: userDefaultsCompletedKey) else { return }

        let numberOfSessions = UserDefaults.standard.value(forKey: "AskForRating") as? Int ?? 0

        if numberOfSessions >= 3 {
            UserDefaults.standard.set(true, forKey: userDefaultsCompletedKey)
            if let scene = UIApplication.shared.currentScene {
                SKStoreReviewController.requestReview(in: scene)
            }
        }
    }
```

julia-andersson (approved):

PR #1221, opened 2024-02-08
fix ChatInputView being cut off/missing padding on iPad

julia-andersson:
## [APP-XXX]

- fix ChatInputView being cut off/missing padding on iPad

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Chat/Sources/Views/ChatScreen.swift
+++ b/Projects/Chat/Sources/Views/ChatScreen.swift
@@ -20,10 +20,9 @@ struct ChatScreen: View {
             infoCard
                 .padding(.bottom, -8)
             ChatInputView(vm: vm.chatInputVm)
-                .padding(.bottom, 32)
+                .padding(.bottom, 16)
         }
         .dismissKeyboard()
-        .edgesIgnoringSafeArea(.bottom)
     }
 
     @ViewBuilder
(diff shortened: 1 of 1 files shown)


Projects/Chat/Sources/Views/ChatScreen.swift
```diff
@@ -20,8 +20,10 @@ struct ChatScreen: View {
             infoCard
                 .padding(.bottom, -8)
             ChatInputView(vm: vm.chatInputVm)
+                .padding(.bottom, 32)
```
sladan-hedvig, line 23: Maybe change this to 16 so its the same as from the side

Projects/Chat/Sources/Views/ChatScreen.swift
```diff
@@ -20,8 +20,10 @@ struct ChatScreen: View {
...
         }
         .dismissKeyboard()
+        .edgesIgnoringSafeArea(.bottom)
```
sladan-hedvig, line 26: .edgesIgnoringSafeArea(.bottom)
When we add this, view doesn't move together with the keyboard 😢 
I think its not needed

sladan-hedvig (changes requested):

sladan-hedvig (approved):

PR #1229, opened 2024-02-14
Update authlib

StylianosGakis:
## [APP-XXX]

- Change 1
Bump Authlib and adjust the code to the new API. Namely the error response type was enhanced to better show what kind of error there was.
- Change 2
There was a breaking change in the schema too, with some input to a mutation going from non-null to nullable, so GQL generates the function acceting an Optional now instead of just the value itself. I've fixed that with wrapping the value with `GraphQLNullable.some()`.
If this was fixed in main already in a different way then I can adjust this too.

### DIFF
--- a/Projects/App/Config/Production/Info.plist
+++ b/Projects/App/Config/Production/Info.plist
@@ -26,7 +26,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>12.4.7</string>
+	<string>12.4.8</string>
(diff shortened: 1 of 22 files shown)


julia-andersson:
iOS developer next! 🍎🔥

sladan-hedvig:
[](https://media.giphy.com/media/xT1XGAeeIvRyYKhkAw/giphy.gif?cid=790b7611xs7eao1r8ii96ug3u0hnujrjbp98xr7078pkfd9w&ep=v1_gifs_search&rid=giphy.gif&ct=g)

PR #1244, opened 2024-02-21
GEN-1854: update claim status card

julia-andersson:
## [GEN-1854]

- [Figma](https://www.figma.com/file/WlwFCAWywXNcXVgPfYVmWu/Final-Design?type=design&node-id=9635-18248&mode=design&t=S79zyH9TZuCSdaes-0)
- [Notion](https://www.notion.so/hedviginsurance/a4c35052b6794b36a4b0d58b537cc325?v=c74a8cb958ab4ba6ba4c904d8540b106&p=1ba2bb29fb3a46f49a84225c65ba41f6&pm=s)
- Added new design and schema for claim status card

- Missing update claims status on backend (submitted/being handled/closed)

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Claims/Sources/Views/ClaimDetailView.swift
```diff
@@ -167,6 +77,175 @@ public struct ClaimDetailView: View {
...
+                .foregroundColor(hTextColor.secondary)
+            Spacer()
+            hText(value)
+                .foregroundColor(hTextColor.secondary)
+        }
+    }
+
+    @ViewBuilder
+    private var termsAndConditions: some View {
+        if let termsAndConditionsDocument = vm.claim.productVariant?.documents.first {
```
sladan-hedvig, line 158: It feels a little bit scary to use first one. Isn't there some better way to get only t&c document?

julia-andersson: Fixed now!

Projects/Contracts/Sources/Models/InsuranceTermModel.swift
sladan-hedvig, line 1: Instad of having :

```
public enum TypeOfDocument: String, Codable {
    case termsAndConditions = "TERMS_AND_CONDITIONS"
    case preSaleInfoEuStandard = "PRE_SALE_INFO_EU_STANDARD"
    case preSaleInfo = "PRE_SALE_INFO"
    case generalTerms = "GENERAL_TERMS"
    case privacyPolicy = "PRIVACY_POLICY"
    case unknown = "UNKNOWN"
    static func resolve(for typeOfDocument: String) -> Self {
        if let concreteTypeOfDocument = Self(rawValue: typeOfDocument) {
            return concreteTypeOfDocument
        }
        log.warn(
            "Got an unknown type of document \(typeOfDocument) that couldn't be resolved.",
            error: nil,
            attributes: nil
        )
        return .unknown
    }
}
```



I think it would be much better to do like this:
```
public enum TypeOfDocument: Codable {
    case termsAndConditions
    case preSaleInfoEuStandard
    case preSaleInfo
    case generalTerms
    case privacyPolicy
    case unknown
}

extension GraphQLEnum<OctopusGraphQL.InsuranceDocumentType> {
    var asTypeOfDocument: TypeOfDocument {
        switch self {
        case .case(let type):
            switch type {
            case .generalTerms:
                return .generalTerms
            case .preSaleInfo:
                return .preSaleInfo
            case .preSaleInfoEuStandard:
                return .preSaleInfoEuStandard
            case .privacyPolicy:
                return .privacyPolicy
            case .termsAndConditions:
                return .termsAndConditions
            }
        case .unknown(let string):
            return .unknown
        }
    }
}

```
and then to have
        self.type = data.type.asTypeOfDocument
instead of 
        self.type = TypeOfDocument.resolve(for: data.type.rawValue)

Why? Because we don't want to initilize TypeOfDocument with octopus type

sladan-hedvig (changes requested):
Left a comment. I hope it makes sense

sladan-hedvig (approved):
I think it looks good now :)

PR #1250, opened 2024-02-26
Add haptic feedback for hRow

julia-andersson:
## [APP-XXX]

- Change 1
- Change 2

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Home/Sources/Screens/HelpCenter/HelpCenterStartView.swift
```diff
@@ -133,13 +139,15 @@ public struct HelpCenterStartView: View {
                         Spacer()
                     }
                     .withChevronAccessory
+                    .onTap {
+                        let generator = UIImpactFeedbackGenerator(style: .light)
```
sladan-hedvig, line 143: Dont need here generator if we are going to use .onTap of hRow which already use generator

sladan-hedvig (approved):
Just that one small changes 👍

PR #1274, opened 2024-03-07
GEN-1877: Added ask for review after successful self service

sladan-hedvig:
- Ask for review after successful self service 

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


julia-andersson (commented):
Should we not get this review pop up for change bank as well?

julia-andersson (approved):

PR #1277, opened 2024-03-08
Bump/authlib

StylianosGakis:
Bumpity bump
I tested locally logging in to NO. And to SE by scanning the moving QR code.
The one core change is that we do not get the liveQrData anymore on the original response, we only start getting it after we start observing it.
But we do start observing it imediatelly after, so practically this should be the same.

The new authlib version removes a bunch of unused stuff like the Callbacks being gone since it was not used anymore. Also it's not an option at all to try to login with ZignSec anymore


StylianosGakis:
I think this will have to rely on the imageUrl not being null being fixed in another PR before CI passes

sladan-hedvig (approved):
Did some updates. Should be awesome! :) noice work!

StylianosGakis:
Thanks so much! I always forget you can just do an extension like
```
extension Localization.Locale.Market {
    fileprivate var asOtpMarket: OtpMarket {
        switch self {
        case .no: return .no
        case .se: return .se
        case .dk: return .dk
        }
    }
}
```
for what I was doing here.
Thanks for fixing this 🤗

StylianosGakis:
Lmk if this brings any other issues, I could not test locally right now due to the aforementioned issues with Xcode 🚧

sladan-hedvig:
> Lmk if this brings any other issues, I could not test locally right now due to the aforementioned issues with Xcode 🚧

It looks nice :) we will see if there are any issues but didn't notice anything

PR #1279, opened 2024-03-08
Improvement/remove  test targets tuist

julia-andersson:
## [APP-XXX]

- Remove examples in tuist that we are not using

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig (approved):
Ok, lets remove them. As we start adding tests to the projects, we can just simply add exampes and tests

PR #1280, opened 2024-03-11
Moved get a price and cross sell to open in safari

sladan-hedvig:
- Open "Get a price quote" and cross sell links directly in safari

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/App/Sources/Journeys/LoggedInJourney.swift
+++ b/Projects/App/Sources/Journeys/LoggedInJourney.swift
@@ -36,10 +36,7 @@ extension AppJourney {
                 case .openCrossSells:
                     CrossSellingScreen.journey { result in
                         if case .openCrossSellingWebUrl(let url) = result {
-                            ContinueJourney()
-                                .onPresent {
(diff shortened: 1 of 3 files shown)


Projects/App/Sources/Journeys/LoggedInJourney.swift
```diff
@@ -36,7 +36,10 @@ extension AppJourney {
                 case .openCrossSells:
                     CrossSellingScreen.journey { result in
                         if case .openCrossSellingWebUrl(let url) = result {
-                            AppJourney.webRedirect(url: url)
+                            ContinueJourney()
+                                .onPresent {
```
julia-andersson, line 40: Would be nice to move this to a separate function, like we have for webRedirect. since it's reused. Maybe "webRedirectInSafari" or "webRedirectInBroswer"

julia-andersson (changes requested):

julia-andersson (approved):

PR #1286, opened 2024-03-13
GEN-1922: prefill model name

julia-andersson:
## [GEN-1922]

- If possible and we have any matching name from availableModels with current device, prefill this under model

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Claims/Sources/Models/FlowClamSingleItemStepModel.swift
+++ b/Projects/Claims/Sources/Models/FlowClamSingleItemStepModel.swift
@@ -37,11 +37,13 @@ public struct FlowClamSingleItemStepModel: FlowClaimStepModel {
         self.selectedItemModel = data.selectedItemModel
 
         if self.selectedItemModel == nil && self.customName == nil {
-            let currentDeviceName = UIDevice.modelName
+            let currentDeviceName = UIDevice.modelName.lowercased()
             if let matchingModelWithCurrentDevice = self.availableItemModelOptions.first(where: {
-                $0.displayName == currentDeviceName
+                let name = $0.displayName.lowercased().trimmingCharacters(in: .whitespacesAndNewlines)
+                return name == currentDeviceName
             }) {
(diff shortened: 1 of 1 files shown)


Projects/Claims/Sources/Models/FlowClamSingleItemStepModel.swift
```diff
@@ -34,6 +35,16 @@ public struct FlowClamSingleItemStepModel: FlowClaimStepModel {
...
+
+        if self.selectedItemModel == nil && self.customName == nil {
+            let currentDeviceName = "Apple " + UIDevice.modelName
+            if let matchingModelWithCurrentDevice = self.availableItemModelOptions.first(where: {
+                $0.displayName == currentDeviceName
+            }) {
+                self.selectedItemModel = matchingModelWithCurrentDevice.itemModelId
+            }
+        }
+
```
sladan-hedvig, line 47: Maybe better
```
            let currentDeviceName = UIDevice.modelName.lowercased()
            if let matchingModelWithCurrentDevice = self.availableItemModelOptions.first(where: {
                let name = $0.displayName.lowercased().replacingOccurrences(of: "apple", with: "").trimmingCharacters(in: .whitespacesAndNewlines)
                return name == currentDeviceName
            }) {
                self.selectedItemModel = matchingModelWithCurrentDevice.itemModelId
                self.selectedItemBrand = matchingModelWithCurrentDevice.itemBrandId
            }
```

sladan-hedvig: Why? When they remove apple we don't have to care since it will still work.

julia-andersson: we don't have apple in the string anymore so we don't need .replacingOccurrences(of: "apple", with: "")

sladan-hedvig (approved):

PR #1287, opened 2024-03-14
Bump authlib and use SKIE to get a better Swift API to the library

StylianosGakis:
- Bump Authlib to the version that uses [SKIE](https://skie.touchlab.co/)
- Remove the polling code which observes the login status and use the provided function by authlib which does it for you
- Try to also migrate TokenRefresher into using the async function, but have less confidence in how I use Signal there, since there was 1 more call site which wanted the Signal version, some interceptor which was checking if there was a need to refresh

I would love to discuss this before we consider merging. Let's do it today in the office

## Checklist

- [x] Tested locally


sladan-hedvig (approved):
Approved! 👌

PR #1296, opened 2024-03-19
Use consistent L10n for tab names

StylianosGakis:
Context:
https://hedviginsurance.slack.com/archives/C05TQ9TFQ6M/p1710850325979789?thread_ts=1710844864.449339&cid=C05TQ9TFQ6M

Should work when CI pulls the latest strings from lokalise.
They live here https://app.lokalise.com/project/743091915e9da969db9340.20943733/?view=multi&filter=tag_fd9160bbd656f75ab74e8935e4330d9e


StylianosGakis:
Changed the keys inside DeepLink as well, however I am not sure how that's used exactly. If it's wrong to change it there too feel free to revert

julia-andersson (approved):

PR #1297, opened 2024-03-19
fill PaymentDetails and change wrong text color

julia-andersson:
## [APP-XXX]

- Fill PaymentDetails with info

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Payment/Sources/Models/PaymentData.swift
+++ b/Projects/Payment/Sources/Models/PaymentData.swift
@@ -78,6 +78,12 @@ public struct PaymentData: Codable, Equatable {
             self.bank = bank
         }
 
+        init(with model: OctopusGraphQL.PaymentInformationQuery.Data) {
+            self.paymentMethod = L10n.paymentsAutogiroLabel
+            self.account = model.currentMember.paymentInformation.connection?.descriptor ?? ""
+            self.bank = model.currentMember.paymentInformation.connection?.displayName ?? ""
+        }
+
         var getDisplayList: [KeyValue] {
             var list: [KeyValue] = []
             list.append((L10n.paymentsPaymentMethod, paymentMethod))
--- a/Projects/Payment/Sources/Models/PaymentData.swift
+++ b/Projects/Payment/Sources/Models/PaymentData.swift
@@ -68,20 +68,14 @@ public struct PaymentData: Codable, Equatable {
 
     struct PaymentDetails: Codable, Equatable {
         typealias KeyValue = (key: String, value: String)
-        private let paymentMethod: String
-        private let account: String
-        private let bank: String
+        let paymentMethod: String
+        let account: String
+        let bank: String
 
         init(paymentMethod: String, account: String, bank: String) {
             self.paymentMethod = paymentMethod
             self.account = account
             self.bank = bank
-        }
-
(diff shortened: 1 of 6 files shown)


Projects/Payment/Sources/Models/PaymentData.swift
```diff
@@ -78,6 +78,12 @@ public struct PaymentData: Codable, Equatable {
             self.bank = bank
         }
 
+        init(with model: OctopusGraphQL.PaymentInformationQuery.Data) {
```
sladan-hedvig, line 81: Can we move this init to octopus implementation?
Also if we can fill this data only if the member has connected payments method

sladan-hedvig (changes requested):

Projects/Payment/Sources/Services/PaymentService/PaymentServiceOctopus.swift
```diff
@@ -11,7 +11,18 @@ public class hPaymentServiceOctopus: hPaymentService {
...
+        let paymentDetailsData = try await octopus.client.fetch(
+            query: paymentDetailsQuery,
+            cachePolicy: .fetchIgnoringCacheCompletely
+        )
+
+        if paymentDetailsData.currentMember.paymentInformation.connection != nil {
+            let paymentDetails = PaymentData.PaymentDetails(with: paymentDetailsData)
+            return PaymentData(with: data, paymentDetails: paymentDetails)
+        }
+        return PaymentData(with: data, paymentDetails: nil)
```
sladan-hedvig, line 25: Instead of this:
```
        if paymentDetailsData.currentMember.paymentInformation.connection != nil {
            let paymentDetails = PaymentData.PaymentDetails(with: paymentDetailsData)
            return PaymentData(with: data, paymentDetails: paymentDetails)
        }
        return PaymentData(with: data, paymentDetails: nil)

```
we can do this
```
let paymentDetails = PaymentData.PaymentDetails(with: paymentDetailsData)
return PaymentData(with: data, paymentDetails: paymentDetails)
```

and instead of this
```
    init(with model: OctopusGraphQL.PaymentInformationQuery.Data) {
        self.paymentMethod = L10n.paymentsAutogiroLabel
        self.account = model.currentMember.paymentInformation.connection?.descriptor ?? ""
        self.bank = model.currentMember.paymentInformation.connection?.displayName ?? ""
    }
```

we can do this
```
    init?(with model: OctopusGraphQL.PaymentInformationQuery.Data) {
        guard let account =  model.currentMember.paymentInformation.connection?.descriptor,
              let bank = model.currentMember.paymentInformation.connection?.displayName else { return nil }
        self.paymentMethod = L10n.paymentsAutogiroLabel
        self.account = account
        self.bank = bank
    }
```
What do you think?
We are skipping having "return PaymentData(" 2 times and init PaymentDetails only when it is needed.

sladan-hedvig (approved):
If you want you can do it. Both ways work.

Projects/Payment/Sources/Services/PaymentService/PaymentServiceOctopus.swift
julia-andersson: yeah make sense

PR #1306, opened 2024-03-25
GEN-1952: New termination flow

julia-andersson:
## [GEN-1952]

- Updated termination flow design
-  Added deeplink support
- Added termination in help center

- To fix:
- Layout looks a bit off on this [screen](https://www.figma.com/file/WlwFCAWywXNcXVgPfYVmWu/Final-Design?type=design&node-id=11129-56288&mode=design&t=agQJReqe6eTvMmCx-0)  

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/hCoreUI/Sources/CheckboxPickerScreen.swift
```diff
@@ -1,8 +1,21 @@
 import SwiftUI
 import hCore
 
+public struct DisplayString: Hashable {
```
sladan-hedvig, line 4: Can we rename this to something more suitable? 
Maybe CheckboxItemModel

Projects/Contracts/Sources/Contracts.swift
```diff
@@ -107,6 +107,14 @@ extension Contracts {
                 )
             } else if case .openContractDetailErrorScreen = action {
                 ContractDetail(id: "", title: "").contractDetailErrorJourney
+            } else if case .openTerminationErrorScreen = action {
+                HostingJourney(
+                    rootView: GenericErrorView(
+                        title: "Couldn'terminate contract",
```
sladan-hedvig, line 113: Can we lokalise this?

sladan-hedvig: And do we need this since we are never calling it?

Projects/Contracts/Sources/Journeys/ContractDetail.swift
```diff
@@ -210,6 +210,30 @@ extension ContractDetail {
             }
         }
     }
+
+    public var terminationErrorJourney: some JourneyPresentation {
```
sladan-hedvig, line 214: We never use this, do we need it?

Projects/Contracts/Sources/Stores/ContractAction.swift
```diff
@@ -38,12 +38,13 @@ public enum ContractAction: ActionProtocol, Hashable {
     case openDetail(contractId: String, title: String)
     case openTerminatedContracts
     case openContractDetailErrorScreen
+    case openTerminationErrorScreen
```
sladan-hedvig, line 41: never used

Projects/hCore/Sources/DeepLink.swift
```diff
@@ -42,6 +43,8 @@ public enum DeepLink: String, Codable {
             return L10n.hcTitle
         case .moveContract:
             return L10n.InsuranceDetails.changeAddressButton
+        case .terminateContract:
+            return "Terminate contract"
```
sladan-hedvig, line 47: Missing lokalise

Projects/hCoreUI/Sources/CheckboxPickerScreen.swift
```diff
@@ -14,21 +27,24 @@ public struct CheckboxPickerScreen<T>: View where T: Equatable & Hashable {
...
     @State private var selectedItems: [T] = []
     @Environment(\.hButtonIsLoading) var isLoading
     @Environment(\.hCheckboxPickerBottomAttachedView) var bottomAttachedView
     @Environment(\.hIncludeManualInput) var includeManualInput
+    @Environment(\.hUseColoredCheckbox) var coloredCheckBox
 
     @State var manualBrandName: String = ""
     @State var manualInput: Bool = false
 
     private var fieldSize: hFieldSize
```
sladan-hedvig, line 44: can we change this to use     
@Environment(\.hFieldSize) var size?

and instead of passing it in init, we can use it 
by  .hFieldSize(.small || .large) after we init CheckboxPickerScreen

Projects/hCoreUI/Sources/hForm/hFloatingField.swift
```diff
@@ -10,6 +10,7 @@ public struct hFloatingField: View {
     @Binding var error: String?
     private var value: String
     private let onTap: () -> Void
+    private let lockedState: Bool
```
sladan-hedvig, line 13: we can remove this here and use 
environment what we have
`    @Environment(\.hFieldLockedState) var isLocked
`

if we add additional extension here
```
extension View {
    public var hFieldLockedState: some View {
        self.environment(\.hFieldLockedState, true)
    }
}
```

with 

```
extension View {
    public var hFieldLockedState: some View {
        self.environment(\.hFieldLockedState, true)
    }
    
    public func hFieldSetLockedState(to value: Bool) -> some View {
        self.environment(\.hFieldLockedState, value)
    }
}

```

sladan-hedvig (changes requested):

Projects/Contracts/Sources/Contracts.swift
julia-andersson: Good catch, I'll remove it for now

Projects/Contracts/Sources/Journeys/ContractDetail.swift
julia-andersson: Yeah same here, will remove it

Projects/hCoreUI/Sources/CheckboxPickerScreen.swift
julia-andersson: I'm not sure this will work since we want to initialize with

```
            if items.count > 3 {
                self.fieldSize = .small
            } else {
                self.fieldSize = .large
            }
```
if we don't get any value on size. One option could be to check this from where we call CheckboxPickerScreen instead but I don't know if that's better than the current solution?

sladan-hedvig: Yeah, ok. Thats fine then :)

julia-andersson: Yeah, we could do this in the future though if we don't need this check anymore :)

sladan-hedvig (approved):

PR #1313, opened 2024-04-03
GEN-1984 GEN-1981 GEN-1982: Change minimum iOS version from 14 to 15

julia-andersson:
## [GEN-1982]

- Change minimum iOS version from 14 to 15

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/App/Sources/Journeys/MainJourney.swift
```diff
@@ -13,6 +13,10 @@ extension AppJourney {
                 UpdateAppScreen.journey.onPresent {
                     Launch.shared.completeAnimationCallbacker.callAll()
                 }
+            } else if Dependencies.featureFlags().osVersionTooLow {
```
sladan-hedvig, line 16: Can we do this like this?
```
 if Dependencies.featureFlags().isUpdateNecessary {
                if Dependencies.featureFlags().osVersionTooLow {
                    UpdateOSScreen.journey.onPresent {
                        Launch.shared.completeAnimationCallbacker.callAll()
                    }
                } else {
                    UpdateAppScreen.journey.onPresent {
                        Launch.shared.completeAnimationCallbacker.callAll()
                    }
                }
            } else {
               ....
            }
```

Why? We only want to show this when we force member to update the app.

Projects/hCoreUI/Sources/Views/MarqueeText.swift
```diff
@@ -154,13 +154,7 @@ extension String {
 extension View {
     /// A backwards compatible wrapper for iOS 14 `onChange`
     @ViewBuilder func onValueChanged<T: Equatable>(of value: T, perform onChange: @escaping (T) -> Void) -> some View {
```
sladan-hedvig, line 156: Can we remove this one, and use it directly where we call "onValueChanged" ?

sladan-hedvig (changes requested):
One small change and one related to the OS version.

PR #1320, opened 2024-04-09
New navigation: home

julia-andersson:
## [APP-XXX]

- Navigation for home.

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Home/Sources/Screens/Home.swift
```diff
@@ -11,10 +11,31 @@ import hCore
...
+    @Published public var isSubmitClaimPresented = false
+    @Published public var isHelpCenterPresented = false
+
+    //claim details
+    @Published public var isChatPresented = false
+    @Published public var isDocumentPresented = false
+    @Published public var document: InsuranceTerm? = nil
+
+    //help center
+    @Published public var isPaymentsPresented = false
```
julia-andersson, line 26: Should these (quick actions) be moved inside its own view model in Help Center instead?

sladan-hedvig (approved):
This to fix:
- MainNavigationJourney should be in charge of handing what root view should be shown - not logged in, logged in, Impersonate - move logic from AppDelegate to MainNavigationJourney when and what we show
- Add navigation for not logged
- Add navigation for logged in
- Add navigation for Impersonate
- Update handling of the datadog views since now its broken
- Revisit help centar navigation logic

PR #1343, opened 2024-05-10
Improvement/navigation/chat

julia-andersson:
## [APP-XXX]

- Rework chat navigation

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig (approved):
Looks good but need to handle links somehow.
We can take that latter on

PR #1349, opened 2024-05-14
Termination survey

sladan-hedvig:
- All changes related to the survey ( and termination flow)
- Updated logic when we start termination flow and which screen we should show - it was always either dateScreen or selectInsurance
- Added hTextView ( messy but it is working)
- Added hUseNewDesign - really getting messy with it
- Added newStep for the survey
- Added handling strange scrollView offset when we present keyboard
- Other UI fixes

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig:
God help us all

PR #1375, opened 2024-06-14
GEN-2172: Add padding variables

julia-andersson:
## [GEN-2172]

- added padding variables
- switched ro use these for some 

## Checklist

- [ ] Landscape verification
- [ ] iPad verification
- [ ]


sladan-hedvig (approved):
Wow, so many changes. Noice!

PR #1399, opened 2024-07-05
Fix/better hanlding of local notificaitons

sladan-hedvig:
- Removed notification toast when on home or chat screen
- Fetch for chat notifications every 10s when on home screen

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


julia-andersson (approved):
Works fine 👍

PR #1414, opened 2024-07-15
remove squircle

julia-andersson:
## [APP-XXX]

- Replace squircle with RoundedRectangle according to new design

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Claims/Sources/Views/SubmitClaim/FilesUpload/FilesUploadScreen.swift
```diff
@@ -56,7 +56,8 @@ struct SubmitClaimFilesUploadScreen: View {
 
                             }
                             .fixedSize(horizontal: false, vertical: true)
-                            .clipShape(Squircle.default())
+                            .background(hSurfaceColor.Opaque.primary)
```
sladan-hedvig, line 59: Do we want to set background here? It wasn't here before

julia-andersson: not sure, I don't think it matters but I can remove it

sladan-hedvig (approved):

PR #1415, opened 2024-07-16
Feature/cbm/update when to show chat

julia-andersson:
## [APP-XXX]

- Update inbox icon to only be visible if there is at least one message

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig (approved):
I think this looks fine. There is one small thing that we can live without - showing chat icon on the home screen for the old chat when we have claim and no messages. Its fine since the member can enter to the chat from claim details. 👌

PR #1422, opened 2024-07-19
File preview update

sladan-hedvig:
- detent file preview
- with updated design
- with swipe left right to dismiss

another version can be found [here](https://github.com/HedvigInsurance/ugglan/pull/1421) 

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Chat/Sources/Views/InboxView.swift
```diff
@@ -41,24 +41,28 @@ public struct InboxView: View {
...
-        VStack(alignment: .leading, spacing: 8) {
-            HStack(alignment: .top) {
-                if conversation.type == .legacy {
-                    hText(L10n.chatConversationHistoryTitle, style: .body1)
-                        .foregroundColor(hTextColor.Opaque.primary)
-                } else {
-                    VStack(alignment: .leading, spacing: 0) {
-                        hText(conversation.title, style: .body1)
+        VStack(alignment: .leading, spacing: 0) {
+            VStack(alignment: .leading, spacing: 0) {
```
julia-andersson, line 45: do we really need duplicate of this VStack?

julia-andersson: maybe we do? or is there a "cleaner" way to do this?

sladan-hedvig: Yeah.... because of the paddings.
After inner VStack we want to add bottom at padding, and outer VStack says that that its a VStack
We could to 
```
```
to 
```
 VStack(alignment: .leading, spacing: 8) {
            VStack(alignment: .leading) {
```
And remove 
`.padding(.bottom, 8)`
Afterwards...

julia-andersson: maybe it's fine then 👍

PR #1441, opened 2024-08-02
Changed alignment of hRow to fix Device ID alignment

danieljohannsson:
## [None]

- Fixed a minor bug where the device ID was top aligned instead of being vertically centered

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/hCoreUI/Sources/hForm/hRow.swift
+++ b/Projects/hCoreUI/Sources/hForm/hRow.swift
@@ -85,7 +85,7 @@ public struct hRow<Content: View, Accessory: View>: View {
     public var body: some View {
         VStack(alignment: .leading, spacing: 0) {
             VStack(alignment: .leading) {
-                HStack(alignment: .top) {
+                HStack(alignment: .center) {
                     content
                     accessory
                 }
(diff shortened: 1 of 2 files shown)


Projects/hCoreUI/Sources/hForm/hRow.swift
```diff
@@ -85,7 +85,7 @@ public struct hRow<Content: View, Accessory: View>: View {
     public var body: some View {
         VStack(alignment: .leading, spacing: 0) {
             VStack(alignment: .leading) {
-                HStack(alignment: .top) {
+                HStack(alignment: .center) {
```
julia-andersson, line 88: Nice thought! I don't think we should not change hRow though since the changes will be reflected everywhere where hRow is used. Instead, we should edit in that affected view (AppInfoView in this case)

julia-andersson: you could for example add `.frame(maxHeight: .infinity, alignment: .center)` to the hText element there 😄

julia-andersson (changes requested):
Nice thought! I don't think we should not change hRow though since the changes will be because

Projects/hCoreUI/Sources/hForm/hRow.swift
julia-andersson: `                hText(deviceId, style: .standardSmall)
                    .minimumScaleFactor(0.2)
                    .lineLimit(1)
                    .foregroundColor(hTextColor.Opaque.secondary)
                    .frame(maxHeight: .infinity, alignment: .center)`

julia-andersson (approved):

PR #1445, opened 2024-08-07
remove title and subtitle from backend and update local logic for it

julia-andersson:
## [APP-XXX]

- Stop query title and subtitle from conversation
- Added logic to handle this locally

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


StylianosGakis (approved):
Yup, looking great

PR #1449, opened 2024-08-20
Update fonts

julia-andersson:
## [APP-XXX]

- Remove old and update with new fonts according to [Figma](https://www.figma.com/design/5kmmDdh6StpXzbEfr7WevV/Hedvig-UI-Kit?node-id=15404-2214&m=dev)

## Checklist
- [ ] Landscape verification
- [ ] iPad verification


Projects/hCoreUI/Sources/hText/hText.swift
```diff
@@ -34,28 +34,39 @@ public enum HFontTextStyle {
...
-    case title1
-    case title2
-    case title3
-    case headline
-    case subheadline
-    case callout
-    case footnote
-    case standardSmall
-    case badge
+    case displayXXLShort
```
julia-andersson, line 49: @sladan-hedvig we could delete these and add later if necessary. What do you think?

sladan-hedvig (approved):
Looks awesome!

PR #1509, opened 2024-09-13
remove chat service type and add list of failed messages

juliaendre:
## [APP-XXX]

- Save failed messages locally when chat is closed

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Chat/Sources/ChatStore.swift
+++ b/Projects/Chat/Sources/ChatStore.swift
@@ -6,13 +6,13 @@ public struct ChatState: StateProtocol {
     public init() {}
 
     @Transient(defaultValue: false) var askedForPushNotificationsPermission: Bool
-    public var failedMessages: [Message] = []
+    public var failedMessages: [String: [Message]] = [:]
 }
 
 public enum ChatAction: ActionProtocol {
     case checkPushNotificationStatus
-    case setFailedMessage(Message)
-    case clearFailedMessage(Message)
+    case setFailedMessage(conversationId: String, message: Message)
+    case clearFailedMessage(conversationId: String, message: Message)
 }
 
 final public class ChatStore: StateStore<ChatState, ChatAction> {
@@ -27,14 +27,16 @@ final public class ChatStore: StateStore<ChatState, ChatAction> {
         switch action {
         case .checkPushNotificationStatus:
             newState.askedForPushNotificationsPermission = true
-        case let .setFailedMessage(message):
+        case let .setFailedMessage(conversationId, message):
             var failedMessages = state.failedMessages
-            failedMessages.append(message)
+            var messages = failedMessages[conversationId]
+            messages?.append(message)
+            failedMessages[conversationId] = messages
             newState.failedMessages = failedMessages
-        case let .clearFailedMessage(message):
+        case let .clearFailedMessage(conversationId, message):
             var failedMessages = state.failedMessages
-            if let index = failedMessages.firstIndex(where: { $0.id == message.id }) {
-                failedMessages.remove(at: index)
+            if let index = failedMessages[conversationId]?.firstIndex(where: { $0.id == message.id }) {
+                failedMessages[conversationId]?.remove(at: index)
             }
             newState.failedMessages = failedMessages
         }
--- a/Projects/Chat/Sources/ChatStore.swift
+++ b/Projects/Chat/Sources/ChatStore.swift
@@ -4,11 +4,15 @@ import hCore
 
 public struct ChatState: StateProtocol {
     public init() {}
+
     @Transient(defaultValue: false) var askedForPushNotificationsPermission: Bool
+    public var failedMessages: [Message] = []
 }
 
 public enum ChatAction: ActionProtocol {
     case checkPushNotificationStatus
+    case setFailedMessage(Message)
+    case clearFailedMessage(Message)
 }
 
 final public class ChatStore: StateStore<ChatState, ChatAction> {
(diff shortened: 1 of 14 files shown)


Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -126,16 +129,20 @@ public class ChatScreenViewModel: ObservableObject {
...
             withAnimation {
                 self.messages.append(contentsOf: newMessages)
+                self.messages.append(contentsOf: store.state.failedMessages)
                 self.messages.sort(by: { $0.sentAt > $1.sentAt })
                 self.lastDeliveredMessage = self.messages.first(where: { $0.sender == .member && $0.remoteId != nil })
             }
             self.banner = chatData.banner
             self.conversationStatus = chatData.conversationStatus ?? .open
             addedMessagesIds.append(contentsOf: newMessages.compactMap({ $0.id }))
+            addedMessagesIds.append(contentsOf: store.state.failedMessages.compactMap({ $0.id }))
```
sladan-hedvig, line 144: No need for this if we do it in the comment before

Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -157,6 +164,8 @@ public class ChatScreenViewModel: ObservableObject {
 
     @MainActor
     func retrySending(message: Message) async {
+        let store: ChatStore = globalPresentableStoreContainer.get()
+        store.send(.clearFailedMessage(message))
```
sladan-hedvig, line 168: I think this cleaning should be done when the message was sent. So no need for this code, lets add it when the message is sent


here:
```
    private func sendToClient(message: Message) async {
        do {
            let sentMessage = try await chatService.send(message: message)
            let store: ChatStore = globalPresentableStoreContainer.get()
            store.send(.clearFailedMessage(message))
```

Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -241,6 +250,8 @@ public class ChatScreenViewModel: ObservableObject {
     private func handleSendFail(for message: Message, with error: String) {
         if let index = messages.firstIndex(where: { $0.id == message.id }) {
             let newMessage = message.asFailed(with: error)
+            let store: ChatStore = globalPresentableStoreContainer.get()
+            store.send(.setFailedMessage(newMessage))
```
sladan-hedvig, line 254: This is maybe placed in the wrong place - we would append new messages again and again. We want to add it once it failed, not if it is already failed. That logic for adding should be moved a little down.
```
            switch oldMessage.status {
            case .failed:
                break
            default:
                messages[index] = newMessage
                let store: ChatStore = globalPresentableStoreContainer.get()
                store.send(.setFailedMessage(newMessage))
            }
```

sladan-hedvig (changes requested):

Projects/Chat/Sources/ChatStore.swift
```diff
@@ -23,6 +27,18 @@ final public class ChatStore: StateStore<ChatState, ChatAction> {
         switch action {
         case .checkPushNotificationStatus:
             newState.askedForPushNotificationsPermission = true
+        case let .setFailedMessage(conversationId, message):
+            var failedMessages = state.failedMessages
+            var messages = failedMessages[conversationId]
```
sladan-hedvig, line 32: Can we replace this with
`var messages = failedMessages[conversationId] ?? []`
Why?
How its written now, we would never have failed messages for the conversation since messages will be nil
Therefore append will never add new message
Therefore failedMessages[conversationId] will be set to nil allways

Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -13,6 +13,7 @@ public class ChatScreenViewModel: ObservableObject {
     @Published var scrollToMessage: Message?
     @Published var banner: Markdown?
     @Published var conversationStatus: ConversationStatus = .open
+    @Published private var conversationId: String = ""
```
sladan-hedvig, line 16: Can we make this optinal? 
No need for published since we will not observe its value anywhere.

Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -108,6 +110,7 @@ public class ChatScreenViewModel: ObservableObject {
                 }
                 self.banner = chatData.banner
                 self.conversationStatus = chatData.conversationStatus ?? .open
+                self.conversationId = chatData.conversationId
```
sladan-hedvig, line 113: No need for this, since we will set conversationId in the `fetchMessages`

Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -126,16 +129,27 @@ public class ChatScreenViewModel: ObservableObject {
...
+                        addedMessagesIds.append(contentsOf: failedMessages.compactMap({ $0.id }))
+                    }
+                }
+
                 self.messages.sort(by: { $0.sentAt > $1.sentAt })
                 self.lastDeliveredMessage = self.messages.first(where: { $0.sender == .member && $0.remoteId != nil })
             }
             self.banner = chatData.banner
             self.conversationStatus = chatData.conversationStatus ?? .open
+            self.conversationId = chatData.conversationId
```
sladan-hedvig, line 150: Can we move this to the line just after
let chatData = try await chatService.getNewMessages() ?
We should set it at the moment we fetch it.

sladan-hedvig (changes requested):

sladan-hedvig (approved):
xD I think its fine now

PR #1511, opened 2024-09-13
CBM: show claims details when clicking on title

juliaendre:
## [APP-XXX]

- Show claim details when clicking on title for claim conversations

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/App/Sources/Journeys/LoggedInNavigation.swift
+++ b/Projects/App/Sources/Journeys/LoggedInNavigation.swift
@@ -100,7 +100,7 @@ struct LoggedInNavigation: View {
         ContractsNavigation(contractsNavigationVm: vm.contractsNavigationVm) { redirectType in
             switch redirectType {
             case .chat:
-                ChatScreen(vm: .init(chatService: NewConversationService()))
+                ChatScreen(vm: .init(chatService: NewConversationService(), onTitleTap: {}))
             case .movingFlow:
                 MovingFlowNavigation()
             case let .pdf(document):
@@ -369,15 +369,26 @@ struct HomeTab: View {
             options: $homeNavigationVm.openChatOptions,
             content: { openChat in
                 ChatNavigation(
-                    chatType: openChat.chatType
-                ) { type, onDone in
-                    AskForPushNotifications(
-                        text: L10n.chatActivateNotificationsBody,
-                        onActionExecuted: {
-                            onDone()
+                    chatType: openChat.chatType,
+                    redirectView: { viewType, onDone in
+                        switch viewType {
+                        case .notification:
+                            AskForPushNotifications(
+                                text: L10n.chatActivateNotificationsBody,
+                                onActionExecuted: {
+                                    onDone()
+                                }
+                            )
+                        case let .claimDetail(id):
+                            let claimStore: ClaimsStore = globalPresentableStoreContainer.get()
+                            if let claim = claimStore.state.claim(for: id) {
+                                ClaimDetailView(claim: claim)
+                                    .configureTitle(L10n.claimsYourClaim)
+                            }
                         }
-                    )
-                }
+                    }
+                )
+
             }
         )
     }
--- a/Projects/App/Sources/Journeys/LoggedInNavigation.swift
+++ b/Projects/App/Sources/Journeys/LoggedInNavigation.swift
@@ -69,7 +69,7 @@ struct LoggedInNavigation: View {
                 homeStore.send(.fetchQuickActions)
             case .chat:
                 DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) {
-                    NotificationCenter.default.post(name: .openChat, object: nil)
+                    NotificationCenter.default.post(name: .openChat, object: ChatType.newConversation)
                 }
             case let .openFeedback(url):
                 vm.openUrl(url: url)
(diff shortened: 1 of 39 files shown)


Projects/Chat/Sources/Models/Conversation.swift
```diff
@@ -12,6 +12,7 @@ public struct Conversation: Identifiable, Hashable, Codable {
         status: ConversationStatus,
         hasClaim: Bool,
         claimType: String?,
+        claimId: String?,
```
sladan-hedvig, line 15: I don't think we need this one. We will fetch it from the ChatViewModel. No need to have it in the conversation itself.

Projects/Chat/Sources/Navigation/ChatNavigation.swift
```diff
@@ -73,8 +84,17 @@ public struct ChatNavigation<Content: View>: View {
         RouterHost(router: router, options: .navigationType(type: .large), tracking: ChatNavigationViewName.chat) {
             Group {
                 switch chatType {
-                case let .conversationId(id):
-                    ChatScreen(vm: .init(chatService: ConversationService(conversationId: id)))
+                case let .conversationId(conversationId, claimId):
```
sladan-hedvig, line 87: Same, I don't think we need to pass claimId around.
ChatScreen model should know whenever there is a claim or not.

Projects/Chat/Sources/Navigation/ChatNavigation.swift
```diff
@@ -44,8 +45,16 @@ public class ChatNavigationViewModel: ObservableObject {
...
 
-public enum ChatRedirectViewType {
+public enum ChatRedirectViewType: Hashable {
     case notification
+    case claimDetail(id: String)
+}
+
+extension ChatRedirectViewType: TrackingViewNameProtocol {
+    public var nameForTracking: String {
+        return "trackingName"
```
sladan-hedvig, line 55: WE can do the proper naming now.
switch self and place proper data - "ClaimDetailView" or "AskForPushNotifications"

Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -28,10 +28,14 @@ public class ChatScreenViewModel: ObservableObject {
     private var isFetching = false
     private var haveSentAMessage = false
     private var openDeepLinkObserver: NSObjectProtocol?
+    private var onTitleTap: () -> Void?
```
sladan-hedvig, line 31: I think this function should send claimId

Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -279,9 +283,16 @@ extension ChatScreenViewModel: TitleView {
...
         let view: UIView = UIHostingController(rootView: titleView).view
         view.backgroundColor = .clear
+        let tapGesture = UITapGestureRecognizer(target: self, action: #selector(handleTapGesture))
+        view.addGestureRecognizer(tapGesture)
+        view.isUserInteractionEnabled = true
         return view
     }
 
+    @objc private func handleTapGesture() {
+        self.onTitleTap()
```
sladan-hedvig, line 293: here we should send claim id. Claim id is id that we fetch from the ChatData

Projects/Home/Sources/Navigation/HomeNavigation.swift
```diff
@@ -33,8 +33,8 @@ public class HomeNavigationViewModel: ObservableObject {
 
             if let chatType = notification.object as? ChatType {
                 switch chatType {
-                case let .conversationId(id):
-                    openChat = .init(chatType: .conversationId(id: id))
+                case let .conversationId(conversationId, claimId):
```
sladan-hedvig, line 36: no need for claim id here. 
We will fetch it from the conversation itself

sladan-hedvig (changes requested):
Can we also hide

Projects/Chat/Sources/Navigation/ChatNavigation.swift
julia-andersson: Uups, totally forgot about this one :)

julia-andersson:
Think I fixed everything now, however I'm struggling a bit with the router/navigation environment object vars 😅 so please if you want to have a look @sladan-hedvig

Projects/Chat/Sources/Navigation/ChatNavigation.swift
```diff
@@ -71,7 +71,7 @@ public enum ChatNavigationViewName: TrackingViewNameProtocol {
 }
 
 public struct ChatNavigation<Content: View>: View {
-    @StateObject var router = Router()
+    @EnvironmentObject var router: Router
```
sladan-hedvig, line 74: I am not sure why we changed this from StateObject to EnvironmentObject.
Any reason to do so?

Projects/App/Sources/Journeys/LoggedInNavigation.swift
```diff
@@ -382,13 +383,14 @@ struct HomeTab: View {
...
-                                ClaimDetailView(claim: claim)
+                                ClaimDetailView(claim: claim, fromChat: true)
                                     .configureTitle(L10n.claimsYourClaim)
+
                             }
                         }
                     }
                 )
-
+                .environmentObject(homeNavigationVm.router)
```
sladan-hedvig, line 393: If we return 
@StateObject var router = Router()
then this is not necessary. Also I think this homeNavigationVm router is router that we use when we want to push something in the home view. 
It would be best to just keep router in the ChatNavigation as an @StateObject

sladan-hedvig (changes requested):
Otherwise looks good.
I think the only think that doesn't feel natural is to clicking on the title.

Projects/Chat/Sources/Navigation/ChatNavigation.swift
juliaendre: Yeah I messed it up a bit :) fixed now!

sladan-hedvig (approved):
Awesome! 🥳

PR #1521, opened 2024-09-23
Presentation store race condition fix

sladan-hedvig:
- It can happen that store isn't saved properly when 2 actions are sent at the same time or close to each other
- Added tests for the home screen to test this

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


juliaendre (approved):
Seems to work fine 🤌

sladan-hedvig:
> Member

Will check a little more since I am afraid it will brake something

PR #1555, opened 2024-10-10
Help center suggested changes for multiple insurances

sladan-hedvig:
what
- added input for list of the contracts
- reverted changeTierInput to only accept contractId

why
- better handling what contract we want to change tier for
- 2 different inputs make nice separation and more control

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/ChangeTier/Sources/Navigation/ChangeTierNavigation.swift
```diff
@@ -12,12 +12,30 @@ public class ChangeTierNavigationViewModel: ObservableObject {
     @Published public var document: Document?
 
     let router = Router()
-    var vm: ChangeTierViewModel
 
+    //make sure to set it!!!
```
juliaendre, line 16: we should remove this

juliaendre (approved):

PR #1560, opened 2024-10-11
Added missing quotes info screeen

sladan-hedvig:
## [APP-XXX]

- Handled from help centar - with or without select insurance screen
- Handled from insurance, edit insurance, upgrade coverage
- Handled from termination flow

From edit insurance:
simulator_screenshot_56E2DA58-9CB1-4DD8-9BC3-8C505B50BA8E

From termination flow:
simulator_screenshot_CAE6FE46-98FF-4116-9E52-9F4628F1FB41

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/ChangeTier/Sources/Navigation/ChangeTierNavigation.swift
+++ b/Projects/ChangeTier/Sources/Navigation/ChangeTierNavigation.swift
@@ -52,9 +52,7 @@ public class ChangeTierNavigationViewModel: ObservableObject {
     }
 
     func missingQuotesGoBackPressed() {
-        if changeTierContractsInput != nil {
-            router.pop()
-        } else if useOwnNavigation {
+        if useOwnNavigation && changeTierContractsInput == nil {
             router.dismiss()
         } else {
             router.pop()
--- a/Projects/ChangeTier/Sources/Navigation/ChangeTierNavigation.swift
+++ b/Projects/ChangeTier/Sources/Navigation/ChangeTierNavigation.swift
@@ -10,7 +10,7 @@ public class ChangeTierNavigationViewModel: ObservableObject {
     @Published public var isCompareTiersPresented = false
     @Published public var isInsurableLimitPresented: InsurableLimits?
     @Published public var document: Document?
-
+    let useOwnNavigation: Bool
     let router: Router
 
     //make sure to set it!!!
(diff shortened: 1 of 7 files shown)


Projects/ChangeTier/Sources/Navigation/ChangeTierNavigation.swift
```diff
@@ -10,7 +10,7 @@ public class ChangeTierNavigationViewModel: ObservableObject {
     @Published public var isCompareTiersPresented = false
     @Published public var isInsurableLimitPresented: InsurableLimits?
     @Published public var document: Document?
-
+    let useOwnNavigation: Bool
     let router: Router
 
     //make sure to set it!!!
```
juliaendre, line 16: can we remove this comment?

Projects/ChangeTier/Sources/Navigation/ChangeTierNavigation.swift
```diff
@@ -41,13 +42,24 @@ public class ChangeTierNavigationViewModel: ObservableObject {
...
+        self.useOwnNavigation = true
     }
 
     public static func getTiers(input: ChangeTierInputData) async throws -> ChangeTierIntentModel {
         let client: ChangeTierClient = Dependencies.shared.resolve()
         let data = try await client.getTier(input: input)
         return data
     }
+
+    func missingQuotesGoBackPressed() {
```
juliaendre, line 54: Is this not the same as?

```
    func missingQuotesGoBackPressed() {
        if useOwnNavigation {
            router.dismiss()
        } else {
            router.pop()
        }
    }
```

I think that would be cleaner and I don't see the purpose of checking `changeTierContractsInput `

juliaendre (changes requested):

Projects/ChangeTier/Sources/Navigation/ChangeTierNavigation.swift
sladan-hedvig: This is like info, can we keep it?

juliaendre: Ah yes but then I think we should rephrase it to look more professional. Maybe "Note: important to set this." or something like that

Projects/ChangeTier/Sources/Navigation/ChangeTierNavigation.swift
sladan-hedvig: Not really since we could use own navigation and still have changeTierContractsInput

sladan-hedvig: But yeah, can be merged 👍

juliaendre (approved):

PR #1563, opened 2024-10-14
Fix for tier displayName showing duplicate values

juliaendre:
## [APP-XXX]

- Change 1
- Change 2

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/ChangeTier/Sources/Models/ChangeTierIntentModel.swift
```diff
@@ -40,83 +44,80 @@ public struct Tier: Codable, Equatable, Hashable, Identifiable {
...
     }
 }
 
-public struct Deductible: Codable, Hashable, Identifiable {
+public struct Quote: Codable, Hashable, Identifiable {
     public var id: String
-
-    let deductibleAmount: MonetaryAmount?
-    let deductiblePercentage: Int?
+    let quoteAmount: MonetaryAmount?
```
sladan-hedvig, line 74: This should still remain deductibleAmoount and deductiblePercentage because it is deductible related to the quote

sladan-hedvig (approved):
We should have
deductibleAmount and deductiblePercentage
instead of quoteAmount and quotePercentage.
Otherwise looks good

PR #1577, opened 2024-10-18
Replace Document with InsuranceTerm to use same one

juliaendre:
## [APP-XXX]

- I have really been confused about those Insurance Terms and Documents since they contain the same thing. I therefore decided to remove Document and only use InsuranceTerm.

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/App/Sources/Journeys/LoggedInNavigation.swift
+++ b/Projects/App/Sources/Journeys/LoggedInNavigation.swift
@@ -114,7 +114,7 @@ struct LoggedInNavigation: View {
             case .movingFlow:
                 MovingFlowNavigation()
             case let .pdf(document):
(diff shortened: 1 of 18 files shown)


sladan-hedvig (commented):
If you ask me, this seems a little bit odd to have PDFPreview initialised with InsurableLimit object and making us that if we want to open PDFPreview we have to use InsurableLimit object. Maybe Document wasn't the best but PdfDocument would do pretty nice, and if needed maybe we can remove InsurableLimits and replace it with PDFDocument since opening PDFPreview with PDFDocument make more sense 😅 Wdyt?

sladan-hedvig (approved):

PR #1590, opened 2024-10-30
Remove store from moving flow

juliaendre:
## [APP-XXX]

- Remove store from moving flow

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/MoveFlow/Sources/View/MovingFlowHousingType.swift
```diff
@@ -74,21 +81,24 @@ public struct MovingFlowHousingTypeView: View {
...
-                    Spacer()
-                    hButton.LargeButton(type: .ghost) {
-                        router.dismiss()
-                    } content: {
-                        hText(L10n.generalCancelButton)
-                    }
-                }
+                .hDisableScroll
+        }
+        .onAppear {
```
sladan-hedvig, line 86: Can we handle this from the view model somehow?
Now we reset addressInputModel every time when this views appear

Projects/MoveFlow/Sources/View/MovingFlowHouseView.swift
```diff
@@ -5,33 +5,23 @@ import hCoreUI
 import hGraphQL
 
 struct MovingFlowHouseView: View {
-    @ObservedObject var vm: HouseInformationInputModel
+    @StateObject var houseInformationInputvm = HouseInformationInputModel()
```
sladan-hedvig, line 8: Can we have this as an observedObject? Now it gets cleaned if we go back to to address view.

sladan-hedvig (changes requested):

sladan-hedvig (approved):

PR #1603, opened 2024-11-12
GEN-2853: Added generic dismiss button

sladan-hedvig:
## [APP-XXX]

- Added dismiss alert for the change tier flow
- Added dismiss alert for the moving flow

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


juliaendre:
Does does only apply for moving flow or can we add it to be used in the other projects as well?

sladan-hedvig:
> Does it only apply for moving flow or can we add it to be used in the other projects as well?

It is added to the **moving flow** and **change tier flow**.
It can be used anywhere.
Decided to keep old dismiss alert for the claims flow.

juliaendre (approved):

PR #1621, opened 2024-11-28
fix titles in claims flow to be according to new design system

juliaendre:
## [APP-XXX]

- Change 1
- Change 2

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig (changes requested):
Can we also change it for:
SelectContractView
SumitClaimEmergencySelectView
?

PR #1637, opened 2024-12-09
Added extraCoverageItems to termination summary

juliaendre:
## [APP-XXX]

- Added extraCoverageItems from schema into model and view

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/TerminateContracts/Sources/Views/TerminationSummaryScreen.swift
+++ b/Projects/TerminateContracts/Sources/Views/TerminationSummaryScreen.swift
@@ -3,9 +3,29 @@ import hCore
 import hCoreUI
 
 struct TerminationSummaryScreen: View {
-    @EnvironmentObject var terminationNavigationVm: TerminationFlowNavigationViewModel
+    @ObservedObject var terminationNavigationVm: TerminationFlowNavigationViewModel
+    let extraCoverageItems: [ExtraCoverageItem]?
+    let withAddonView: Bool
 
-    init() {}
+    init(
+        terminationNavigationVm: TerminationFlowNavigationViewModel
+    ) {
+        self.terminationNavigationVm = terminationNavigationVm
+
+        self.extraCoverageItems = {
+            let dateExtraCoverageItems = terminationNavigationVm.terminationDateStepModel?.extraCoverageItem
+            let deletionExtraCoverageItems = terminationNavigationVm.terminationDeleteStepModel?.extraCoverageItem
+
+            if let dateExtraCoverageItems, !dateExtraCoverageItems.isEmpty {
+                return dateExtraCoverageItems
+            } else if let deletionExtraCoverageItems, !deletionExtraCoverageItems.isEmpty {
+                return deletionExtraCoverageItems
+            }
+            return nil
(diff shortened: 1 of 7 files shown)


Projects/TerminateContracts/Sources/Views/TerminationSummaryScreen.swift
```diff
@@ -86,60 +102,38 @@ struct TerminationSummaryScreen: View {
     private var addonContent: (some View)? {
         VStack(alignment: .leading, spacing: 0) {
             hText(L10n.terminationAddonCoverageTitle)
-            ForEach(terminationNavigationVm.config?.addonDisplayItems ?? [], id: \.self) { item in
+            ForEach(extraCoverageItems ?? [], id: \.self) { item in
```
sladan-hedvig, line 105: just use then `terminationNavigationVm.extraCoverage` here

Projects/TerminateContracts/Sources/Views/TerminationSummaryScreen.swift
```diff
@@ -3,9 +3,29 @@ import hCore
...
+    let extraCoverageItems: [ExtraCoverageItem]?
+    let withAddonView: Bool
 
-    init() {}
+    init(
+        terminationNavigationVm: TerminationFlowNavigationViewModel
+    ) {
+        self.terminationNavigationVm = terminationNavigationVm
+
+        self.extraCoverageItems = {
```
sladan-hedvig, line 15: Can we have this logic inside TerminationFlowNavigationViewModel ?
like:
`    @Published var extraCoverage: [ExtraCoverageItem] = []
`
and just update
```
    @Published var terminationDateStepModel: TerminationFlowDateNextStepModel? {
        didSet{
            extraCoverage = terminationDateStepModel?.extraCoverageItem ?? []
        }
    }
    
    @Published var terminationDeleteStepModel: TerminationFlowDeletionNextModel?{
        didSet{
            extraCoverage = terminationDeleteStepModel?.extraCoverageItem ?? []
        }
    }
```

Which should enable you to remove
```
    let extraCoverageItems: [ExtraCoverageItem]?
    let withAddonView: Bool
```
here

Projects/TerminateContracts/Sources/Views/TerminationSummaryScreen.swift
```diff
@@ -86,60 +102,38 @@ struct TerminationSummaryScreen: View {
     private var addonContent: (some View)? {
```
sladan-hedvig, line 102: No need for `(some View)?` here - simply `some View` would do.
Also
`    @ViewBuilder`
is unnecessary since we are returning VStack, not something that can vary

sladan-hedvig (changes requested):
Just placed some comments to remove some logic to view model instead to have it in the view

juliaendre:
@sladan-hedvig Adressed your changes. I still need to have withAddonView to determine if we should have bottom view in the status card, but I moved it to the view instead

sladan-hedvig (approved):
Looks good 👍

PR #1640, opened 2024-12-10
Feature/addons/add select insurance screen fixes suggestions

sladan-hedvig:
## [APP-XXX]

- Did some fixes related to the new flow

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Addons/Sources/Navigation/ChangeAddonNavigation.swift
```diff
@@ -29,7 +29,7 @@ class ChangeAddonNavigationViewModel: ObservableObject {
     @Published var isChangeCoverageDaysPresented: AddonOffer?
     @Published var isConfirmAddonPresented = false
     @Published var isAddonProcessingPresented = false
-    @Published var changeAddonVm: ChangeAddonViewModel
+    @Published var changeAddonVm: ChangeAddonViewModel?
```
sladan-hedvig, line 32: Added it as an optional since we can have changeAddonNavigation with list of contracts so we don't want to init it with empty contractId

Projects/Addons/Sources/Views/AddonSelectInsuranceScreen.swift
```diff
@@ -1,10 +1,11 @@
+import Combine
 import SwiftUI
 import hCore
 import hCoreUI
 
 public struct AddonSelectInsuranceScreen: View {
```
sladan-hedvig, line 6: Updated this View so now we track loading and progress

Projects/Addons/Sources/Views/ChangeAddonSummaryScreen.swift
```diff
@@ -49,9 +50,6 @@ extension ChangeAddonViewModel {
             ),
             isAddon: true
         ) {
-            Task {
```
sladan-hedvig, line 52: Removed this from here it was placed wrongly. It was calling submitAddons when we press "Add  and confirm" instead on bottom sheet on click on "Yes, upgrade my coverage".

Projects/Addons/Sources/Views/ConfirmChangeAddonScreen.swift
```diff
@@ -25,6 +25,9 @@ public struct ConfirmChangeAddonScreen: View {
                             hButton.LargeButton(type: .primary) {
                                 addonNavigationVm.isAddonProcessingPresented = true
                                 addonNavigationVm.isConfirmAddonPresented = false
+                                Task {
+                                    await addonNavigationVm.changeAddonVm!.submitAddons()
```
sladan-hedvig, line 29: Moved here since we want to call it when member confirms it!

juliaendre (approved):
Makes sense 🤌

PR #1642, opened 2024-12-12
Schema implementation for addons

juliaendre:
## [APP-XXX]

- Added query for `upsellTravelAddonBanner` from travel insurance and insurances
- Added mutations `upsellTravelAddonOffer` and `upsellTravelAddonActivate`
- Excluded: AddonVariant

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Addons/GraphQL/Octopus/CurrentAddonContract.graphql
```diff
@@ -0,0 +1,6 @@
+query CurrentAddonContract($contractId: UUID!) {
```
juliaendre, line 1: not sure that we even need this one at all but will leave it until we get AddonVariant

PR #1663, opened 2025-01-13
GEN-3124: Display addon perils

sladan-hedvig:
- Display addon perils in insurance details coverage

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/hCoreUI/Sources/Views/CoverageView.swift
+++ b/Projects/hCoreUI/Sources/Views/CoverageView.swift
@@ -5,12 +5,12 @@ import hGraphQL
 public struct CoverageView: View {
     let limits: [InsurableLimits]
     var didTapInsurableLimit: (_ limit: InsurableLimits) -> Void
-    let perils: [Perils]
+    let perils: [(title: String?, perils: [Perils])]
 
     public init(
         limits: [InsurableLimits],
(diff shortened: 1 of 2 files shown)


Projects/hCoreUI/Sources/Views/CoverageView.swift
```diff
@@ -24,10 +24,48 @@ public struct CoverageView: View {
             ) { limit in
                 didTapInsurableLimit(limit)
             }
-            PerilCollection(
-                perils: perils
-            )
-            .hFieldSize(.small)
+            VStack(spacing: .padding32) {
+                ForEach(perils, id: \.title) { perils in
```
panasetskaya, line 28: still no idea about the swift:) so just writing a random thought - is it guaranteed here that the main insurance perils will be displayed first, and addons' second, by the order of adding them to the hashmap in allPerils?

panasetskaya (approved):

Projects/hCoreUI/Sources/Views/CoverageView.swift
sladan-hedvig: allPerils return a list of tuples, so its guaranteed 👍

PR #1692, opened 2025-01-23
GEN-3188: Voiceover accessibility

juliaendre:
## [GEN-3188]

- Accessibility voice over
- Probably missed some places but I think this is a good start, at least it's much better than it was before

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Home/Sources/Screens/HelpCenter/ReusableComponents.swift
```diff
@@ -113,11 +113,14 @@ struct SupportView: View {
     var body: some View {
         hSection {
             VStack(spacing: 0) {
-                hText(L10n.hcChatQuestion)
-                    .foregroundColor(hTextColor.Translucent.primary)
-                hText(L10n.hcChatAnswer)
-                    .foregroundColor(hTextColor.Translucent.secondary)
-                    .multilineTextAlignment(.center)
+                VStack {
```
sladan-hedvig, line 116: Can we specify spacing here? It will use default 8 one

Projects/hCoreUI/Sources/Views/InfoCard/InfoCard.swift
```diff
@@ -17,12 +17,23 @@ public struct InfoCard: View {
     }
 
     public var body: some View {
+        if #available(iOS 18.0, *) {
+            mainView
+                .accessibilityHint((buttonsConfig?.first?.buttonTitle ?? ""), isEnabled: buttonsConfig?.count == 1)
+        } else {
+            // Fallback on earlier versions
+            mainView
+        }
```
sladan-hedvig, line 26: Maybe is better to have it like:
```
        if buttonsConfig?.count == 1 {
            mainView.accessibilityHint((buttonsConfig?.first?.buttonTitle ?? ""))
        } else {
            mainView
        }
```

So we support earlier versions.

Projects/hCoreUI/Sources/Views/NotificationStyle.swift
```diff
@@ -1,5 +1,6 @@
 import SwiftUI
 import UIKit
+import hCore
```
sladan-hedvig, line 3: Do we need hCore?

sladan-hedvig (changes requested):
Looks great! I like the ideas with text for buttons depending on selected value.

Put some comments, also noticed that:
- Can’t open text field on termination survey 
- Termination date field reads odd. It includes arrow pointing down
- Summary screen can’t select documents
- Looks like inbox and chat haven't been checked

Projects/hCoreUI/Sources/Views/QuoteSummaryScreen.swift
```diff
@@ -265,12 +265,12 @@ public struct QuoteSummaryScreen: View {
         proxy: ScrollViewProxy,
         isExpanded: Bool
     ) -> some View {
-        if isExpanded {
-            bottomContent(for: contract, proxy: proxy, isExpanded: isExpanded)
-                .accessibilityElement(children: .combine)
-        } else {
-            bottomContent(for: contract, proxy: proxy, isExpanded: isExpanded)
-        }
+        //        if isExpanded {
```
sladan-hedvig, line 268: remove comments

juliaendre: uuups

sladan-hedvig (approved):
Added small change related to the initial language if the member has different language on system and in the app.

Otherwise its good! We can later on work on trying to fix something like number 1 000 000 having milion instead of one zero zero zero zero zero zero xD

PR #1701, opened 2025-01-28
Feature/addons/add addon perils

sladan-hedvig:
## [APP-XXX]

- Updated perils for addon

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/hCore/Sources/Perils.swift
+++ b/Projects/hCore/Sources/Perils.swift
@@ -31,7 +31,7 @@ public struct Perils: Codable, Equatable, Hashable, Sendable {
         title = fragment.title
         description = fragment.description ?? ""
         covered = [fragment.coverageText].compactMap({ $0 })
-        color = nil
+        color = fragment.colorCode
         isDisabled = false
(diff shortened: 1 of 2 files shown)


juliaendre:
simulator_screenshot_74AD84C5-E4D6-4CD4-BB04-18F4AE914BC8

The spacing between the title and the text here looks a bit off (too much) without the colored peril circle 😅 I think we should check if we have image or not and adjust spacing if we don't

juliaendre (changes requested):
Missing the colors 😢

sladan-hedvig:
> simulator_screenshot_74AD84C5-E4D6-4CD4-BB04-18F4AE914BC8
> 
> The horizontal spacing between the title and the text here looks a bit off (too much) without the colored peril circle 😅 I think we should check if we have image or not and adjust spacing if we don't

Added the colors, but they are off. They will update them next week, this is just to ship it by the end of this week.

juliaendre (approved):
🐰

PR #1707, opened 2025-01-31
Rework state view

juliaendre:
## [APP-XXX]

- Reworking state view

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/hCoreUI/Sources/Views/ProcessingStateView.swift
```diff
@@ -102,11 +121,13 @@ public struct ProcessingStateView: View {
...
-                buttons: .init(
+                successViewBody: successViewBody ?? ""
+            )
+            .hStateViewButtonConfig(
+                .init(
                     actionButton: nil,
-                    primaryButton: nil,
-                    ghostButton: .init(buttonAction: successViewButtonAction ?? {})
+                    actionButtonAttachedToBottom: nil,
+                    dismissButton: .init(buttonAction: successViewButtonAction ?? {})
```
sladan-hedvig, line 130: Default button title is L10n.openChat
as here 
 hText(dismissButton.buttonTitle ?? L10n.openChat, style: .body1) inside StateView
If we dont define title here, it will be "Write to us".
Primary idea for this hErrorViewButtonConfig was just for error view. In order to make this better, I think we should check if everywhere is used proper text when defining dismissButton 😕 

Can we define it? I think it should be "Close"
simulator_screenshot_C7F8FE39-EC3B-494B-ADFC-3284DD5A67D4

Projects/hCoreUI/Sources/Views/ProcessingStateView.swift
```diff
@@ -25,7 +44,7 @@ public struct ProcessingStateView: View {
     var showSuccessScreen: Bool
 
     @Environment(\.hCustomSuccessView) var customSuccessView
-    @Environment(\.hErrorViewButtonConfig) var errorViewButtonConfig
+    @Environment(\.hStateViewButtonConfig) var errorViewButtonConfig
```
sladan-hedvig, line 47: No need to have this here since it is not used in this file (it is used just to pass it down here):
```
    @ViewBuilder
    private func errorView(errorMessage: String?) -> some View {
        GenericErrorView(
            title: L10n.somethingWentWrong,
            description: errorMessage,
            formPosition: .center
        )
        .hStateViewButtonConfig(errorViewButtonConfig)
    }
```
But there is no need to pass it since it will be taken from StateView that is used in GenericErrorView.

sladan-hedvig (changes requested):

Projects/hCoreUI/Sources/Views/ProcessingStateView.swift
juliaendre: Great catch! Will fix this

sladan-hedvig (approved):
I checked all things that I could thing of. Looks good :) 
Nice initiative

PR #1718, opened 2025-02-06
Updated router to check for pushed route before being installed on the view

sladan-hedvig:
- Router now check already pushed routes after showing the router view
- Issue was noticed when starting the app from the https://hedvigtest.page.link/contact-info deeplink
## Checklist

- [ ] Landscape verification
- [ ] iPad verification


juliaendre (approved):
works fine!

PR #1731, opened 2025-02-13
GEN-3291: New design for suggestion cards in termination survey

juliaendre:
## [GEN-3291]

- Updated design in termination survey flow for suggestions (info cards)
- We want to use both campaign (green) and info cards (blue) and have with and without buttons
- [Figma](https://www.figma.com/design/zItETkT4mu5ANVDYED5gEm/Retention-UX-Improvements-App?node-id=27-4697&t=iDNAaZcU3PVEKdCR-0)

##Blockers
- Waiting for graphQL implementation on backend

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/TerminateContracts/Sources/Views/TerminationSurveyScreen.swift
+++ b/Projects/TerminateContracts/Sources/Views/TerminationSurveyScreen.swift
@@ -64,7 +64,7 @@ struct TerminationSurveyScreen: View {
                                 .zIndex(1)
                                 Group {
                                     if let suggestion = option.suggestion, option.id == vm.selectedOption?.id {
-                                        buildInfo(for: suggestion)
+                                        suggestionView(for: suggestion)
                                             .matchedGeometryEffect(id: "buildInfo", in: animationNamespace)
                                     }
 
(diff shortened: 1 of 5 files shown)


Projects/TerminateContracts/Sources/Views/TerminationSurveyScreen.swift
```diff
@@ -62,36 +62,33 @@ struct TerminationSurveyScreen: View {
...
-                                    }
-                                }
-                                .frame(minHeight: 120)
                             }
                         }
 
                     }
+                    if let suggestion = vm.selectedOption?.suggestion {
+                        suggestionView(for: suggestion)
+                            .matchedGeometryEffect(id: "buildInfo", in: animationNamespace)
```
sladan-hedvig, line 71: No need for
` .matchedGeometryEffect(id: "buildInfo", in: animationNamespace)
`
 if we have only 1

Projects/TerminateContracts/Sources/Views/TerminationSurveyScreen.swift
```diff
@@ -62,36 +62,33 @@ struct TerminationSurveyScreen: View {
...
         .sectionContainerStyle(.transparent)
     }
 
     @ViewBuilder
-    func buildInfo(for suggestion: TerminationFlowSurveyStepSuggestion) -> some View {
+    func suggestionView(for suggestion: TerminationFlowSurveyStepSuggestion) -> some View {
         switch suggestion {
         case .action(let action):
-            InfoCard(text: action.description, type: .campaign)
+            InfoCard(text: action.description, type: action.type == .offer ? .campaign : .info)
```
sladan-hedvig, line 91: for this one instead of 
InfoCard(text: action.description, type: `action.type == .offer ? .campaign : .info`)
We could have 
InfoCard(text: action.description, type: `action.type.notificationType`)

and extend enum (need to import hCoreUI also)
```
enum SurveySuggestionInfoType: Codable {
    case info
    case offer
    
    var notificationType: NotificationType {
        switch self {
        case .info:
            return .info
        case .offer:
            return .campaign
        }
    }
}
```
so we can reuse it in the other places

sladan-hedvig (approved):
Just those small 'upgrades' but looks nice 
Awesome 🥳

PR #1739, opened 2025-02-18
Fixed issue where force logout doesn't dismiss presented VCs

sladan-hedvig:
## [APP-XXX]

- If we have presented modal, it will be not dismissed if the member get force logout

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/App/Sources/AppDelegate.swift
```diff
@@ -215,4 +216,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             }
         }
     }
+
+    private func dissmissAllVC() {
```
juliaendre, line 220: Typo: dismiss

juliaendre (changes requested):

juliaendre (approved):

PR #1740, opened 2025-02-19
Log purchased and upgraded addons event

sladan-hedvig:
## [APP-XXX]

- Added logs for purchased action
- Added logs for upgraded action

Event can be checked [here](https://app.datadoghq.eu/rum/sessions?query=%40type%3Aaction%20%40application.id%3A416e8fc0-c96a-4485-8c74-84412960a479%20env%3Adev%20%40action.type%3Acustom&agg_m=count&agg_m_source=base&agg_t=count&cols=&fromUser=true&viz=stream&from_ts=1740043227362&to_ts=1740129627362&live=true).

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Addons/Sources/Navigation/ChangeAddonNavigation.swift
+++ b/Projects/Addons/Sources/Navigation/ChangeAddonNavigation.swift
@@ -23,9 +23,9 @@ public struct ChangeAddonInput: Identifiable, Equatable {
 }
 
 public enum AddonSource: String, Codable {
-    case insurances
-    case travelCertificates
-    case deeplink
+    case insurances = "INSURANCES"
+    case travelCertificates = "TRAVEL_CERTIFICATES"
+    case deeplink = "DEEPLINK"
 
     public var getSource: OctopusGraphQL.UpsellTravelAddonFlow {
         switch self {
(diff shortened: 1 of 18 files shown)


Projects/Addons/Sources/Navigation/ChangeAddonNavigation.swift
```diff
@@ -8,29 +8,29 @@ public struct ChangeAddonInput: Identifiable, Equatable {
     public var id: String = UUID().uuidString
 
     let contractConfigs: [AddonConfig]?
-    let addonId: String?
-
+    let addonSource: AddonSource
     public init(
-        contractConfigs: [AddonConfig]? = nil,
-        addonId: String? = nil
+        addonSource: AddonSource,
```
juliaendre, line 13: Maybe we could add `AddonSource` as a property to `AddonConfig` so we just have to send in `AddonConfig` as an argument here? What do you think?

sladan-hedvig: Since AddonConfig is a info about contract and we can have multiple contracts, I opted to have it as a seperated argument.

juliaendre (approved):

PR #1757, opened 2025-03-03
GEN-3365: Improvements/move shared models from graphQl to hCore

sladan-hedvig:
## [APP-XXX]

- Change 1
- Change 2

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Forever/Sources/Service/DemoImplementation/ForeverClientDemo.swift
```diff
@@ -1,3 +1,5 @@
+import hCore
```
juliaendre, line 1: Do we really need hCore here?

juliaendre (approved):

Projects/Forever/Sources/Service/DemoImplementation/ForeverClientDemo.swift
sladan-hedvig: Looks like now, but odd since we use init thats from hCore 😅

PR #1758, opened 2025-03-03
Fetch memberId from home instead of passing it

juliaendre:
## [APP-XXX]

- Fetch memberId from home instead of passing it

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Home/Sources/HomeState.swift
+++ b/Projects/Home/Sources/HomeState.swift
@@ -10,7 +10,7 @@ import hCoreUI
 
 public struct HomeState: StateProtocol {
     public var memberContractState: MemberContractState = .loading
-    public var memberId: String = ""
+    public var memberId: String?
     public var futureStatus: FutureStatus = .none
     public var contracts: [HomeContract] = []
     public var importantMessages: [ImportantMessage] = []
--- a/Projects/Home/Sources/HomeState.swift
+++ b/Projects/Home/Sources/HomeState.swift
@@ -10,6 +10,7 @@ import hCoreUI
 
 public struct HomeState: StateProtocol {
     public var memberContractState: MemberContractState = .loading
+    public var memberId: String = ""
     public var futureStatus: FutureStatus = .none
     public var contracts: [HomeContract] = []
     public var importantMessages: [ImportantMessage] = []
@@ -56,6 +57,7 @@ public enum HomeAction: ActionProtocol {
     case fetchImportantMessages
     case setImportantMessages(messages: [ImportantMessage])
     case setMemberContractState(state: MemberContractState, contracts: [HomeContract])
+    case setMemberId(id: String)
     case setFutureStatus(status: FutureStatus)
     case openDocument(contractURL: URL)
     case fetchQuickActions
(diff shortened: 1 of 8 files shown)


Projects/Home/Sources/HomeState.swift
```diff
@@ -10,6 +10,7 @@ import hCoreUI
 
 public struct HomeState: StateProtocol {
     public var memberContractState: MemberContractState = .loading
+    public var memberId: String = ""
```
sladan-hedvig, line 13: Can we have this as optional sting? It should be either nil or proper value

Projects/Home/Sources/Screens/HomeScreen.swift
```diff
@@ -131,9 +127,11 @@ class HomeVM: ObservableObject {
     @Published var toolbarOptionTypes: [ToolbarOptionType] = []
     private var chatNotificationPullTimer = Timer.publish(every: 10, on: .main, in: .common).autoconnect()
 
-    init(memberId: String) {
-        homeBottomScrollViewModel = .init(memberId: memberId)
+    init() {
         let store: HomeStore = globalPresentableStoreContainer.get()
+        store.send(.fetchMemberState)
+        let memberId = store.state.memberId
```
sladan-hedvig, line 133: This member id could be wrong in the home store when we init this view model. Since we are using this memberId just in homeBottomScrollViewModel for delete requests, its fine.

sladan-hedvig (changes requested):
The reason why we pass it from the profile store was because we had this id for sure in the profile store. It can happen that memberId in the HomeStore is "" when we init HomeVM.

Since memberId is used only for homeBottomScrollViewModel, maybe extract memberId in that model and on update of that memberId, just call handleDeleteRequests

And also make memberId optional in the HomeState

Projects/Home/Sources/Screens/HomeBottomScrollView.swift
```diff
@@ -62,15 +62,17 @@ class HomeBottomScrollViewModel: ObservableObject {
     }
     private var showConnectPaymentCardView = false
     var cancellables = Set<AnyCancellable>()
-    init(memberId: String) {
+    var memberIdObserver = Set<AnyCancellable>()
```
sladan-hedvig, line 65: We can add it to the existing cancellables property. But yeah, thats nice how you handled it xD

Projects/Home/Sources/Screens/HomeScreen.swift
```diff
@@ -131,9 +127,9 @@ class HomeVM: ObservableObject {
     @Published var toolbarOptionTypes: [ToolbarOptionType] = []
     private var chatNotificationPullTimer = Timer.publish(every: 10, on: .main, in: .common).autoconnect()
 
-    init(memberId: String) {
-        homeBottomScrollViewModel = .init(memberId: memberId)
+    init() {
         let store: HomeStore = globalPresentableStoreContainer.get()
+        homeBottomScrollViewModel = .init()
```
sladan-hedvig, line 132: I think that
`private(set) var homeBottomScrollViewModel: HomeBottomScrollViewModel
`
can be changed now to
`let homeBottomScrollViewModel = HomeBottomScrollViewModel()
`

sladan-hedvig (approved):
Minor things, hope they make sense.

Projects/Home/Sources/Screens/HomeBottomScrollView.swift
juliaendre: Do you want me to add it to the existing one instead? :)

sladan-hedvig: Sorry, didn't see comment. We can remove `memberIdObserver` and just use `cancellables` in other places

PR #1763, opened 2025-03-04
Group horizontal padding to hWithoutHorizontalPadding

juliaendre:
## [APP-XXX]

- Group withoutHorizontalPadding for section, row and divider

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Contracts/Sources/View/ContractCoverage.swift
+++ b/Projects/Contracts/Sources/View/ContractCoverage.swift
@@ -24,7 +24,7 @@ struct ContractCoverageView: View {
                     },
                     perils: contract.allPerils
                 )
-                .hRowWithoutHorizontalPadding
+                .hWithoutHorizontalPadding([.row])
             }
         }
     }
--- a/Projects/Contracts/Sources/View/ContractDetail.swift
+++ b/Projects/Contracts/Sources/View/ContractDetail.swift
@@ -84,7 +84,6 @@ public struct ContractDetail: View {
                     )
                     .padding(.top, .padding16)
                     .padding(.bottom, .padding8)
-                    .hWithoutDividerPadding
                 }
                 .sectionContainerStyle(.transparent)
                 .padding(.top, .padding8)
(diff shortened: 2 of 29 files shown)


Projects/Contracts/Sources/View/ContractCoverage.swift
```diff
@@ -24,7 +24,7 @@ struct ContractCoverageView: View {
                     },
                     perils: contract.allPerils
                 )
-                .hRowWithoutHorizontalPadding
+                .hWithoutHorizontalPadding([.row, .divider])
```
sladan-hedvig, line 27: It used to be just a row, and you added divider now. Is that in purpose?

Projects/Contracts/Sources/View/ContractDetail.swift
```diff
@@ -84,7 +84,6 @@ public struct ContractDetail: View {
                     )
                     .padding(.top, .padding16)
                     .padding(.bottom, .padding8)
-                    .hWithoutDividerPadding
```
sladan-hedvig, line 87: I guess this one is moved to contractCoverage

Projects/hCoreUI/Sources/ItemPickerScreen.swift
```diff
@@ -447,8 +446,8 @@ struct ItemPickerScreen_Previews: PreviewProvider {
                         .asAnyView
                 }
             )
-            .hEmbeddedHeader
-            .hIncludeManualInput
+            //            .hEmbeddedHeader
```
sladan-hedvig, line 449: Is this on purpose? If it is, then just remove comments

sladan-hedvig (commented):
The only thing is not having it on the same place, but I guess you test how it looks like. There are some commented code that maybe shouldn't be comment or removed if it should be.

Also interesting and nice approach 👍

sladan-hedvig (approved):

PR #1768, opened 2025-03-06
Added suggestions for moving flow refactor

sladan-hedvig:
## [APP-XXX]

I had something like this in mind:
Having different models for intent and for request and handling selection inside navigation

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


juliaendre:
I'm not sure I think that this is super clear either, when to use request model and when to use intent model. Could we perhaps rename them to something better?

juliaendre:
Yes that's better I think! :)

juliaendre (approved):
Just that renaming and I think it will be fine

PR #1769, opened 2025-03-07
Feature/cross sell own project

juliaendre:
## [APP-XXX]

- Separate cross sell to it's own module since we will be working a lot with cross sell in upcoming projects
- Not sure if we want to fetch cross sell data from service or if we want to keep it in contract store/create cross sell store

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


juliaendre:
For some reason CrossSellingScreen which we call on home doesn't expand to all the content 😅

sladan-hedvig (approved):
I think this is great. Will open window for the new X-sell implementation

PR #1784, opened 2025-03-12
Make submit claims its own module

juliaendre:
## [APP-XXX]

- Separate submit claim from claims

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Claims/Sources/Views/ClaimFilesView.swift
```diff
@@ -237,12 +237,19 @@ class ClaimFilesViewModel: ObservableObject {
         }
     }
 
-    struct ClaimFilesViewOptions: OptionSet {
-        let rawValue: UInt
+    @MainActor
+    public struct ClaimFilesViewOptions: @preconcurrency OptionSet {
```
sladan-hedvig, line 241: `    public struct ClaimFilesViewOptions: OptionSet, Sendable {
`
should be better to use here

+ `@MainActor` can be removed

sladan-hedvig (commented):
Just this minor comment, haven't noticed anything that might cause issues.

sladan-hedvig (approved):
🥳 accidentally commented instead approving

PR #1785, opened 2025-03-12
GEN-3411: Added preselected contract id for contract step in ClaimFlow + refactor ItemPickerScreen

sladan-hedvig:
## [APP-XXX]

- Added preselected contract id
- Refactored ItemPickerScreen since it didn't save data when view dissapear
- Changed it everywhere where is needed + checked all places where its updated 🥳 

Issue was related to that every time ItemPickerScreen is rendered, new ItemConfig is created. 
Having it as a property on the view is solution where it is generated once

Its a lot

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Addons/Sources/Navigation/ChangeAddonNavigation.swift
+++ b/Projects/Addons/Sources/Navigation/ChangeAddonNavigation.swift
@@ -165,7 +165,7 @@ public struct ChangeAddonNavigation: View {
     }
 
     private var selectInsuranceScreen: some View {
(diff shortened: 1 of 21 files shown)


sladan-hedvig:
> Awesome! We should also rework that ItemPickerScreen sometime 🤪

Totally agree! It came across my thoughts also. Like lets push it with this now and see to rework it. It got out of the hand for one simple ui component

PR #1786, opened 2025-03-13
GEN-3432: Enable landscape mode

juliaendre:
## [GEN-3432]

- For accessibility purposes

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig:
I don't like view modifier for orientation, but rather would have observableObject like
import SwiftUI
import Combine

```
@MainActor
class OrientationDetector: ObservableObject {
    @Published var orientation: DeviceOrientation
    private var orientationCancellable: AnyCancellable?
    init(){
        orientation = UIDevice.current.orientation.getDeviceOrientation
        orientationCancellable = NotificationCenter.default.publisher(for: UIDevice.orientationDidChangeNotification).sink {[weak self] _ in
            self?.orientation = UIDevice.current.orientation.getDeviceOrientation
        }
    }
}
public enum DeviceOrientation {
    case portrait
    case landscape
}

private extension UIDeviceOrientation {
    var getDeviceOrientation: DeviceOrientation {
        switch self {
        case .landscapeLeft, .landscapeRight:
            return .landscape
        default:
            return .portrait
        }
    }
}

```
and then 
`    @StateObject private var orientationDetector = OrientationDetector()
`inside hForm
and access orientation by `orientationDetector.orientation`

On this way we don't have to set state everytime when want to use orientation.
And also no need to connect that orientation binding
`.detectOrientation($orientation)
`

sladan-hedvig (approved):
Do we only want this enabled only on Ugglan or should we go for both Ugglan and Hedvig?
I think it looks amazing and we should go with also on the production.

juliaendre:
Oh I just forgot to add it! Will add it for Hedvig as well 😄

PR #1787, opened 2025-03-13
Fix issue with pending insurances

sladan-hedvig:
## [APP-XXX]

- Added card info when having pending contract
- Updated logic about home screen when we have only pending contracts
- Issue was reported [here](https://hedviginsurance.slack.com/archives/CPCUHMMJQ/p1741855953083179)

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig:
> Great that you fixed this!

Well I messed it up also 😅

PR #1788, opened 2025-03-14
Rework item picker screen and add generic contract select step

juliaendre:
## [APP-XXX]

- Refactoring of ItemPickerScreen
- Added ContractSelectView to be reused as contract select step. In this was we can easily update design for all of these screens.

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig (approved):
It looks promising :) 

Did one change CoInsuredSelectInsuranceScreen where itemPickerConfig is generated once instead of every time we render view

PR #1790, opened 2025-03-18
GEN-3387: Claim details push notifications and deep link

juliaendre:
## [GEN-3387]

- Added handling of push notifications `CLAIM_CLOSED` and `OPEN_CLAIM`
- Added handling of deep link `/claim-details?claimId=`

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/App/Sources/Navigation/LoggedInNavigation.swift
```diff
@@ -928,6 +935,25 @@ class LoggedInNavigationViewModel: ObservableObject {
         }
     }
 
+    func handleClaimDetails(url: URL? = nil, claimId: String? = nil) {
+        if let url = url, let claimId = url.getParameter(property: .claimId) {
```
sladan-hedvig, line 939: Can we just pass claimId here? And handle getting `claimId` out of the url before?
And merge openClaimDetails with this function also? 
And make it private?

Projects/App/Sources/Navigation/LoggedInNavigation.swift
```diff
@@ -928,6 +935,25 @@ class LoggedInNavigationViewModel: ObservableObject {
...
+            openClaimDetails(for: claimId)
+        } else if let claimId {
+            openClaimDetails(for: claimId)
+        }
+    }
+
+    private func openClaimDetails(for claimId: String) {
+        let claimStore: ClaimsStore = globalPresentableStoreContainer.get()
+        if let claim = claimStore.state.claim(for: claimId) {
```
sladan-hedvig, line 948: Can we be sure to get the latest claims before we continue here by fetching it? There might be missing claim with specific id in the claimStore.

Projects/App/Sources/Navigation/LoggedInNavigation.swift
```diff
@@ -928,6 +935,25 @@ class LoggedInNavigationViewModel: ObservableObject {
...
+            UIApplication.shared.getRootViewController()?.dismiss(animated: true)
+            self.selectedTab = 0
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) { [weak self] in
```
sladan-hedvig, line 951: I think we should start getting away from using DispatchQueue.main.asyncAfter if possible and just use 
```
Task {[weak self] in
   try await Task.sleep
   self?.homeNavigationVm.router.push(claim)
}
```

sladan-hedvig (changes requested):

Projects/App/Sources/Navigation/LoggedInNavigation.swift
```diff
@@ -928,6 +939,20 @@ class LoggedInNavigationViewModel: ObservableObject {
         }
     }
 
+    private func handleClaimDetails(claimId: String) {
+        let claimStore: ClaimsStore = globalPresentableStoreContainer.get()
+        claimStore.send(.fetchClaims)
```
sladan-hedvig, line 944: If we keep it like this, it will be executed async.
We want to call it with  `wait claimStore.sendAsync(.fetchClaims)` so we wait for this action to be executed.

Also make sure that .setClaims action is also executed with `await sendAsync` inside ClaimsStore

Projects/App/Sources/Navigation/LoggedInNavigation.swift
```diff
@@ -928,6 +939,20 @@ class LoggedInNavigationViewModel: ObservableObject {
...
+
+        if let claim = claimStore.state.claim(for: claimId) {
+            UIApplication.shared.getRootViewController()?.dismiss(animated: true)
+            self.selectedTab = 0
+            Task { [weak self] in
+                try await Task.sleep(nanoseconds: 20_000_000)
```
sladan-hedvig, line 950: 20_000_000 = 0.02 s
dont we want 200_000_000 ?

Projects/App/Sources/Navigation/LoggedInNavigation.swift
```diff
@@ -663,6 +663,13 @@ class LoggedInNavigationViewModel: ObservableObject {
                 Task {
                     await handleTravelAddon()
                 }
+            case .OPEN_CLAIM, .CLAIM_CLOSED:
+                UIApplication.shared.getRootViewController()?.dismiss(animated: true)
```
sladan-hedvig, line 667: Can we move `UIApplication.shared.getRootViewController()?.dismiss(animated: true) ` inside  
```
    private func handleClaimDetails(claimId: String) {
        if let claim = claimStore.state.claim(for: claimId) {
           //here
```
since we want to dismiss it only when we have all necessary data in order to show it

sladan-hedvig (changes requested):
I hope it all makes sense.

Projects/App/Sources/Navigation/LoggedInNavigation.swift
juliaendre: I actually moved it there, just forgot to remove it from here also

sladan-hedvig (approved):
I think its great! 🥳

PR #1791, opened 2025-03-18
Improvement/co insured select insurance screen subtitle

sladan-hedvig:
## [APP-XXX]

- Added subtitle in select contract in edit co insured flow

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig:
> Nice! I think we should add subtitle for TravelInsurance contract select step as well
> 
> Also changed fieldsize since it was too big now when we had subtitle

Yeah, nice. Much better with that size

PR #1792, opened 2025-03-18
GEN-3389: Handle cross sell after dismissing claim details

juliaendre:
## [GEN-3389]

- Added `showClaimClosedFlow` and `claimAcknowledgeClosedStatus` mutation from schema
- After dismissing claim details screen, show cross sell if `showClaimClosedFlow` is true, then call `claimAcknowledgeClosedStatus`

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Claims/Sources/Service/OctopusImplementation/FetchClaimDetailsClientOctopus.swift
+++ b/Projects/Claims/Sources/Service/OctopusImplementation/FetchClaimDetailsClientOctopus.swift
@@ -50,8 +50,8 @@ public class FetchClaimDetailsClientOctopus: hFetchClaimDetailsClient {
         }
     }
 
-    public func acknowledgeClosedStatus(statusId: String) async throws {
-        let mutation = OctopusGraphQL.ClaimAcknowledgeClosedStatusMutation(claimAcknowledgeClosedStatusId: statusId)
+    public func acknowledgeClosedStatus(claimId: String) async throws {
+        let mutation = OctopusGraphQL.ClaimAcknowledgeClosedStatusMutation(claimAcknowledgeClosedStatusId: claimId)
         let data = try await octopus.client.perform(mutation: mutation)
         if let userError = data.claimAcknowledgeClosedStatus?.userError {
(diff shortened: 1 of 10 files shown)


Projects/Claims/Sources/Views/ClaimDetailView.swift
```diff
@@ -136,6 +136,11 @@ public struct ClaimDetailView: View {
                 })
             )
         )
+        .onDisappear {
+            Task {
+                await vm.acknowledgeClosedStatus()
```
sladan-hedvig, line 141: I don't think that onDisappear is a way to do it. This can happen in different occasions, like if we send app to the background. 

I think we shouldn't have it here, but rather here:
```
        ClaimDetailView(claim: claim, type: type)
            .configureTitle(L10n.claimsYourClaim)
            .onDisappear {
                let claimsStore: ClaimsStore = globalPresentableStoreContainer.get()
                if claim?.showClaimClosedFlow ?? false {
                    homeNavigationVm.navBarItems.isNewOfferPresented = true
                    //send acknowledgeClosedStatus here using depende
                    let service: hFetchClaimDetailsClient = Dependencies.shared.resolve()
                    await service.acknowledgeClosedStatus(statusId: claim.id)
                    claimsStore.send(.fetchClaims)
                }
            }
```
Or maybe even using FetchClaimDetailsService, but make it public which I think is not preferable. Probably there it would need additional changes in the code 😅 
and use `onDeinit` instead `onDisappear`

Projects/Claims/Sources/Service/OctopusImplementation/FetchClaimDetailsClientOctopus.swift
```diff
@@ -47,4 +49,12 @@ public class FetchClaimDetailsClientOctopus: hFetchClaimDetailsClient {
             throw FetchClaimDetailsError.noClaimFound
         }
     }
+
+    public func acknowledgeClosedStatus(statusId: String) async throws {
```
sladan-hedvig, line 53: Can we rename attribute to `claimId`?

sladan-hedvig (approved):

PR #1811, opened 2025-03-25
Make Campaign its own module

juliaendre:
## [APP-XXX]

- Change 1
- Change 2

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/Campaign/Sources/CampaignNavigation.swift
```diff
@@ -0,0 +1,113 @@
...
+        redirect: @escaping (CampaignRedirectType) -> Void,
+        onEditCode: @escaping () -> Void
+    ) {
+        self.campaignNavigationVm = campaignNavigationVm
+        self.redirect = redirect
+        self.onEditCode = onEditCode
+    }
+
+    public var body: some View {
+        RouterHost(router: campaignNavigationVm.router, tracking: CampaignRouterAction.discounts) {
```
sladan-hedvig, line 35: I think we shouldn't have RouterHost here since we don't need navigation here.

Projects/Campaign/Sources/CampaignNavigation.swift
```diff
@@ -0,0 +1,113 @@
...
+@MainActor
+public class CampaignNavigationViewModel: ObservableObject {
+    @Published public var isAddCampaignPresented = false
+    @Published public var isDeleteCampaignPresented: Discount?
+    var paymentDataDiscounts: [Discount]
+    public let router = Router()
+
+    public init(paymentDataDiscounts: [Discount]) {
+        self.paymentDataDiscounts = paymentDataDiscounts
+    }
```
sladan-hedvig, line 16: Since we need router passed from whenever we are calling this view, we should pass Router.
   ```
public let router = Router()

public init(paymentDataDiscounts: [Discount]) {
     self.paymentDataDiscounts = paymentDataDiscounts
}
```
    
```
public let router: Router

public init(paymentDataDiscounts: [Discount], router: Router) {
    self.paymentDataDiscounts = paymentDataDiscounts
    self.router = router
}
```

File: Projects/Campaign/Sources/CampaignNavigation.swift [hunk 3]
```diff
@@ -0,0 +1,113 @@
...
+                    discount: discount,
+                    paymentDataDiscounts: campaignNavigationVm.paymentDataDiscounts,
+                    onInputChange: onEditCode
+                )
+            )
+            .embededInNavigation(
+                options: .navigationType(type: .large),
+                tracking: CampaignDetentActions.deleteCampaign
+            )
+        }
```
sladan-hedvig, line 76: WE need one more modifier to handle push for forever
```
            .routerDestination(for: CampaignRouterAction.self) { type in
                ForeverNavigation(useOwnNavigation: false)
                    .hideToolbar()
            }
```
also we need to import Forever

Projects/Campaign/Sources/CampaignNavigation.swift
```diff
@@ -0,0 +1,113 @@
...
+    var paymentDataDiscounts: [Discount]
+    public let router = Router()
+
+    public init(paymentDataDiscounts: [Discount]) {
+        self.paymentDataDiscounts = paymentDataDiscounts
+    }
+}
+
+public struct CampaignNavigation: View {
+    let redirect: (CampaignRedirectType) -> Void
```
sladan-hedvig, line 20: I think we can remove this redirect and just handle that redirection in CampaignNavigation

Projects/Campaign/Sources/CampaignNavigation.swift
```diff
@@ -0,0 +1,113 @@
...
+        case .addCampaign:
+            return .init(describing: AddCampaignCodeView.self)
+        case .deleteCampaign:
+            return .init(describing: DeleteCampaignView.self)
+        }
+    }
+}
+
+public enum CampaignRouterAction: Hashable {
+    case discounts
```
sladan-hedvig, line 95: Since we don't use this one anymore, we can remove it, but we will need .forever one

Projects/Campaign/Sources/CampaignNavigation.swift
```diff
@@ -0,0 +1,113 @@
...
+extension CampaignRouterAction: TrackingViewNameProtocol {
+    public var nameForTracking: String {
+        switch self {
+        case .discounts:
+            return .init(describing: PaymentsDiscountsRootView.self)
+        }
+    }
+}
+
+public enum CampaignRedirectType: Hashable {
```
sladan-hedvig, line 107: WE can remove this one.

Projects/Campaign/Sources/Views/PaymentsDiscountsView.swift
```diff
@@ -7,9 +7,10 @@ import hCoreUI
 
 struct PaymentsDiscountsView: View {
     let data: PaymentDiscountsData
-    @PresentableStore var store: PaymentStore
-    @EnvironmentObject var paymentsNavigationVm: PaymentsNavigationViewModel
+    @PresentableStore var store: CampaignStore
+    @EnvironmentObject var campaignNavigationVm: CampaignNavigationViewModel
     @EnvironmentObject var router: Router
+    let redirect: (CampaignRedirectType) -> Void
```
sladan-hedvig, line 13: WE don't need this since we will use push on router instead.
CampaignRouterAction.Forever

Projects/Payment/Sources/Navigation/PaymentNavigation.swift
```diff
@@ -8,8 +9,6 @@ public class PaymentsNavigationViewModel: ObservableObject {
 
     public init() {}
     public var connectPaymentVm = ConnectPaymentViewModel()
-    @Published public var isAddCampaignPresented = false
-    @Published public var isDeleteCampaignPresented: Discount?
 }
 
 public struct PaymentsNavigation<Content: View>: View {
```
sladan-hedvig, line 14: I think if we remove redirect we can ditch this Content thing

sladan-hedvig (changes requested):
I think it would be better to just remove that redirect and use different approach with router.

sladan-hedvig (approved):

PR #1851, opened 2025-04-30
Improvement for certificate: make UI reusable

juliaendre:
## [APP-XXX]

- Make the common screens for the certificate use the same UI to make it more maintainable

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


juliaendre:
The reason why I thought it would be a good idea is because we changed the design multiple times and it feels like we are often stuck with different designs since we only update one of them

PR #1861, opened 2025-05-09
GEN-3712: contact info card

juliaendre:
## [GEN-3712]

- Change 1
- [Figma](https://www.figma.com/design/zItETkT4mu5ANVDYED5gEm/UX-Improvements--Small-Projects-?node-id=853-19287&p=f&t=Lqvh9X2xlifjgZ0U-0)

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Profile/Sources/Views/Screens/MyInfo/MyInfoView.swift
+++ b/Projects/Profile/Sources/Views/Screens/MyInfo/MyInfoView.swift
@@ -11,111 +11,67 @@ public struct MyInfoView: View {
     public init() {}
 
     public var body: some View {
-        hForm {
-            hSection {
-                VStack(spacing: .padding4) {
-                    hFloatingTextField(
-                        masking: .init(type: .digits),
-                        value: $vm.phone,
-                        equals: $vm.type,
-                        focusValue: .phone,
-                        placeholder: L10n.phoneNumberRowTitle,
-                        error: $vm.phoneError
-                    )
-                    hFloatingTextField(
-                        masking: .init(type: .email),
-                        value: $vm.email,
-                        equals: $vm.type,
-                        focusValue: .email,
-                        placeholder: L10n.emailRowTitle,
-                        error: $vm.emailError
-                    )
+        hForm {}
+            .hFormAttachToBottom {
+                hSection {
+                    VStack(spacing: .padding16) {
+                        VStack(spacing: .padding4) {
+                            infoCardView
+                            emailField
+                            phoneNumberField
+                        }
+                        buttonView
+                    }
                 }
+                .sectionContainerStyle(.transparent)
                 .disabled(vm.isLoading)
             }
-            .padding(.top, .padding8)
-        }
-        .hFormAttachToBottom({
-            hSection {
-                hButton.LargeButton(type: .primary) {
-                    Task {
-                        withAnimation {
-                            vm.isLoading = true
-                        }
-                        await vm.save()
-                        withAnimation {
-                            vm.isLoading = false
-                            vm.checkForChanges()
-                        }
-                    }
-                } content: {
-                    hText(L10n.generalSaveButton)
-                }
-                .hButtonIsLoading(vm.isLoading)
-            }
-            .sectionContainerStyle(.transparent)
-            .padding(.bottom, .padding8)
-            .opacity(vm.inEditMode ? 1 : 0)
-        })
-        .sectionContainerStyle(.transparent)
-        .introspect(.viewController, on: .iOS(.v13...)) { vc in
-            self.vm.vc = vc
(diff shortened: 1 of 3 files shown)


Projects/Home/Sources/Navigation/HomeNavigation.swift
```diff
@@ -92,3 +92,14 @@ public class HomeNavigationViewModel: ObservableObject {
         existingCoInsured: globalPresentableStoreContainer.get(of: ContractStore.self)
     )
 }
+
+public enum HomeRouterActions: TrackingViewNameProtocol {
```
sladan-hedvig, line 96: I don't think we need new RouterActions, but rather something some other way to notify loggedInNavigation to open contactinfo.

Projects/Profile/Sources/Views/Screens/MyInfo/MyInfoView.swift
```diff
@@ -11,111 +10,69 @@ public struct MyInfoView: View {
...
-                    }
-                    .foregroundColor(hTextColor.Opaque.primary)
-                    .opacity(vm.inEditMode ? 1 : 0)
-                    .transition(.opacity.animation(.easeInOut(duration: 0.2)))
-                    .disabled(!vm.inEditMode)
+    private var buttonView: some View {
+        hButton.LargeButton(type: .primary) {
+            Task {
+                withAnimation {
+                    vm.isLoading = true
```
sladan-hedvig, line 63: I think we should ditch this logic around bool and isLoading and just switch to ProcessingState

sladan-hedvig (commented):
There is a PR where we have new mutation and also property to check if we should show info or not. Lets wait for that one and see what is needed to be done to make it work

Projects/Profile/Sources/Views/Screens/MyInfo/MyInfoView.swift
```diff
@@ -126,14 +82,21 @@ public class MyInfoViewModel: ObservableObject {
...
-    @Published var showAlert: Bool = false
-    @Published var isLoading: Bool = false
+    @Published var viewState: ProcessingState = .success
+    @Published var disabledSaveButton: Bool = false
 
-    weak var vc: UIViewController?
     private var originalPhone: String
     private var originalEmail: String
     private var cancellables = Set<AnyCancellable>()
+    var showInfoCard: Bool {
```
panasetskaya, line 91: tbh I left that card to be there at all times 😅  Do you think we should hide it? What about the cases where the contact info is more than 1 year old (aka `isContactInfoUpdateNeeded` is true)?

Projects/Profile/Sources/Views/Screens/MyInfo/MyInfoView.swift
```diff
@@ -144,29 +107,34 @@ public class MyInfoViewModel: ObservableObject {
...
+                self?.isValid()
             }
             .store(in: &cancellables)
     }
 
-    func checkForChanges() {
+    func isValid() {
         withAnimation {
-            inEditMode = originalPhone != phone || originalEmail != email
+            disabledSaveButton = (phone == "" || originalPhone == "") || (email == "" || originalEmail == "")
```
panasetskaya, line 123: So if let's say originalPhone is empty, would the save button be disabled anyway? *reading unfamiliar language noises*

Projects/hGraphQL/GraphQL/Octopus/Profile/MemberUpdateContactInfo.graphql
```diff
@@ -0,0 +1,8 @@
+mutation MemberUpdateContactInfo($input: MemberUpdateContactInfoInput!) {
+  memberUpdateContactInfo(input: $input) {
+    member {
```
panasetskaya, line 3: we also added a userError there, usually the message in it would be null, but in one case there would come a message from BE - if the email is already taken (Sladan said it came up a couple of times in bug_inbox)

panasetskaya (approved):

Projects/Profile/Sources/Views/Screens/MyInfo/MyInfoView.swift
juliaendre: Also realized that I should only use `isContactInfoUpdateNeeded` to check this here, will update it :)

Projects/hGraphQL/GraphQL/Octopus/Profile/MemberUpdateContactInfo.graphql
juliaendre: Great catch! Will add this

panasetskaya (approved):
🚀

PR #1866, opened 2025-05-13
Rework images

juliaendre:
## [APP-XXX]

- Change 1
- Change 2

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/App/Sources/Screens/ClaimsScreen/Screens/HonestyPledge.swift
```diff
@@ -46,37 +46,43 @@ struct SlideToConfirm: View {
                     .frame(maxWidth: .infinity)
                     .opacity(updateUIForFinished ? 0 : (1 - Double(progress / width)))
             }
-            Image(uiImage: updateUIForFinished ? hCoreUIAssets.checkmark.image : hCoreUIAssets.chevronRight.image)
```
sladan-hedvig, line 49: There is a reason why it was initially placed like this. Animating image of the Image view is smoother then replacing whole image view. Check how it looks before and after changes

Projects/TerminateContracts/Sources/Views/TerminationSummaryScreen.swift
```diff
@@ -69,12 +69,9 @@ struct TerminationSummaryScreen: View {
     @ViewBuilder
     private var mainContent: some View {
         HStack(spacing: .padding12) {
-            Image(
-                uiImage: terminationNavigationVm.config?.typeOfContract?.pillowType.bgImage
-                    ?? hCoreUIAssets.pillowHome.image
-            )
-            .resizable()
-            .frame(width: 48, height: 48)
+            terminationNavigationVm.config?.typeOfContract?.pillowType.bgImage
```
sladan-hedvig, line 72: Need to have  default value if we don't get image

Projects/hCoreUI/Sources/Toasts.swift
```diff
@@ -80,11 +81,7 @@ public struct ToastBarView: View {
 
     @hColorBuilder
     private var iconColor: some hColor {
-        if toastModel.icon != nil {
-            hSignalColor.Green.element
-        } else {
-            toastModel.type.imageColor
-        }
+        hSignalColor.Green.element
```
sladan-hedvig, line 84: This function is returning always one value. Just set it in the place where its called.

Tuist/ProjectDescriptionHelpers/Project+Templates.swift
```diff
@@ -242,7 +241,7 @@ extension Project {
                 options: .options(coverage: true, codeCoverageTargets: ["\(name)"])
             )
         }
-        let shouldDisableGeneratingAccessors = !targets.contains(.generateAssets)
+        let shouldDisableGeneratingAccessors = true
```
sladan-hedvig, line 244: Just place it inside where it is used since it is always true

sladan-hedvig (changes requested):
Placed some comments.

Projects/App/Sources/Screens/ClaimsScreen/Screens/HonestyPledge.swift
juliaendre: Ok, will add it back here

sladan-hedvig (approved):
🥳

PR #1869, opened 2025-05-20
GEN-3747: Don’t show connect payment field if data is “-"

juliaendre:
## [GEN-3747]

- UI fix for when `displayName` and `descriptor` has value `-`
- Not sure if we want to have logic like this on client side or if we should get it like null value from BE

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig:
I think we should address this to BE guys so we don't have to clean up this in the app.

PR #1871, opened 2025-05-27
Design fix: Make image appear instant

juliaendre:
## [APP-XXX]

- [Figma](https://www.notion.so/hedviginsurance/Laggy-transition-on-the-image-should-should-appear-instantly-instead-of-sliding-in-1f3c3f71cb0a80e5b263d506530477d1)

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig:
Added some opacity animation since we don't want to just appears there with a BOOM.

PR #1873, opened 2025-05-27
Rework intro video

juliaendre:
## [APP-XXX]

Not sure this is the way we want to do it 😅

- [Notion](https://www.notion.so/hedviginsurance/Intro-video-is-laggy-low-frame-rate-1f3c3f71cb0a802e8330d659a9c791e9)
- Reworked the loading and playing of the image
- Updated the video to have a bit higher quality

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/App/Sources/Views/LoginVideoView.swift
+++ b/Projects/App/Sources/Views/LoginVideoView.swift
@@ -13,8 +13,6 @@ struct LoginVideoView: UIViewRepresentable {
 
 private class PlayerUIView: UIView {
     private let playerLayer = AVPlayerLayer()
-    private var queuePlayer: AVQueuePlayer?
-    private var looper: AVPlayerLooper?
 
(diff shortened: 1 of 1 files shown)


sladan-hedvig:
Size of video is actually smaller and I like that 👍 I used to have logic to reverse video when it reaches end so its nicer, but if we don't want it.. its fine to remove it

sladan-hedvig:
> Size of video is actually smaller and I like that 👍 I used to have logic to reverse video when it reaches end so its nicer, but if we don't want it.. its fine to remove it

@juliaendre

sladan-hedvig:
> Hmm not sure what we want to keep here. What do you think? @sladan-hedvig

I am fine with this if WIliam wants it like this. I just find it nicer to have it reversed when it reaches end of video

sladan-hedvig (approved):

PR #1882, opened 2025-06-02
Added additional layer for smoother scrolling exp

sladan-hedvig:
## [APP-XXX]
- Updated scrolling logic for both scroll view with and without pager
- TODO: merge scrolls into one component

[Notion](https://www.notion.so/hedviginsurance/Scrolling-animation-is-laggy-and-doesn-t-stick-1f3c3f71cb0a808693ecfc34b0b6cebc)

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/hCoreUI/Sources/Views/ScrollableSegmentedView.swift
+++ b/Projects/hCoreUI/Sources/Views/ScrollableSegmentedView.swift
@@ -303,37 +303,43 @@ extension ScrollableSegmentedViewModel: UIScrollViewDelegate {
             }
             var scrollTo: CGFloat = 0
             var useAnimationEarlier = false
+            let offsets = getPagesOffset()
+            let maxIndex = pageModels.count - 1
             if let currentIndex = pageModels.firstIndex(where: { $0.id == currentId }) {
-                if velocity.x > 1 && currentIndex != pageModels.count - 1 {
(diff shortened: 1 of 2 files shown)


Projects/hCoreUI/Sources/Views/ScrollableSegmentedView.swift
```diff
@@ -301,12 +301,31 @@ extension ScrollableSegmentedViewModel: UIScrollViewDelegate {
             if #available(iOS 17.4, *) {
                 scrollView.stopScrollingAndZooming()
             }
-            let scrollTo = getNearestTabOffset(for: targetContentOffset.pointee.x)
+            var scrollTo: CGFloat = 0
+            var useAnimationEarlier = false
+            if let currentIndex = pageModels.firstIndex(where: { $0.id == currentId }) {
```
juliaendre, line 306: Could this part be optimized more? Maybe something like

```
        guard let currentIndex = pageModels.firstIndex(where: { $0.id == currentId }) else { return }

        let offsets = getPagesOffset()
        let maxIndex = pageModels.count - 1

        let targetIndex: Int? = {
            if velocity.x > 1, currentIndex < maxIndex {
                return currentIndex + 1
            } else if velocity.x < -1, currentIndex > 0 {
                return currentIndex - 1
            }
            return nil
        }()
```

juliaendre (approved):
Looks great! But would be nice with some code optimization if possible 🤌

PR #1896, opened 2025-06-12
Feature/cross sell/navigation button animation

sladan-hedvig:
- added rotation animation
This is not the way we want to handle it tbh, but this part require some more work to make it nicer. #legacy

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig:
Will prolong it

PR #1908, opened 2025-06-17
Make custom blurred background

juliaendre:
## [APP-XXX]

- Make custom blurred background based on the background color in figma
- [Notion link](https://www.notion.so/hedviginsurance/Opaque-frosting-bottom-sheet-modal-1f3c3f71cb0a801eb44ddfae2eb401ca)

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/hCoreUI/Sources/PresentationStyle+Detented.swift
+++ b/Projects/hCoreUI/Sources/PresentationStyle+Detented.swift
@@ -448,7 +448,6 @@ public enum Detent: Equatable {
         }
     }
 }
-//}
 
 @available(iOS 16.0, *)
 public class BlurredSheetPresenationController: UISheetPresentationController {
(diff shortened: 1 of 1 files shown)


Projects/hCoreUI/Sources/PresentationStyle+Detented.swift
```diff
@@ -461,7 +460,7 @@ public class BlurredSheetPresenationController: UISheetPresentationController {
         useBlur: Bool
     ) {
         super.init(presentedViewController: presentedViewController, presenting: presentingViewController)
-        effectView = useBlur ? PassThroughEffectView(effect: UIBlurEffect(style: getBlurEffectStyle)) : nil
```
juliaendre, line 464: If we still want to use a default UIBlurEffect we could use something like `systemUltraThinMaterial` but I think this is a better approach

juliaendre: I tried playing around with custom UI View but it just didn't look good 😅 so going for this approach for now

sladan-hedvig (approved):
🥳

PR #1922, opened 2025-06-26
GEN-3893: Feature/text view update

sladan-hedvig:
- update according to the design
- make text view scrollable
- there is some strange behaviour when opening text view from termination flow  probably because of alwaysVisibleBottomView - so I removed hero transition there
- iPad crashes fix (on iPadOS text view automatically becomes first responder so it was causing big issues on iPad)

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/TerminateContracts/Sources/Views/TerminationSurveyScreen.swift
+++ b/Projects/TerminateContracts/Sources/Views/TerminationSurveyScreen.swift
@@ -45,34 +45,32 @@ struct TerminationSurveyScreen: View {
 
     private var content: some View {
         hSection {
-            VStack(spacing: 16) {
-                VStack(spacing: 4) {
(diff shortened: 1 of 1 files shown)


sladan-hedvig:
> Simulator.Screen.Recording.-.iPhone.16.-.2025-06-27.at.10.20.07.mov 
> When clicking on the options with input fields the screen is not scrolled all the way to the bottom. I'm not really sure what the expected behaviour should be but it feels a bit chopped off right now, right?

Yeah, for some reason it moves. It is keyboard related for sure. Decided to leave it. Usually after a week or so it just strikes me why it is happening 😅

sladan-hedvig:
> I see! Should we wait to merge it then?

Yeah, it was happening on the main branch also, so I would treat it as a separated issue.

PR #1924, opened 2025-06-27
GEN-3900: iOS26 auth updates

juliaendre:
## [GEN-3900]

- Needed to make a few adjustment to auth in order to be compatible with iOS26

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/App/Sources/Service/OctopusClientsImplementation/AuthenticationClientAuthLib.swift
+++ b/Projects/App/Sources/Service/OctopusClientsImplementation/AuthenticationClientAuthLib.swift
@@ -6,16 +6,6 @@ import Foundation
 import hCore
 import hGraphQL
 
-protocol HttpClientEngine {
-    func send(request: URLRequest) async throws -> (Data, URLResponse)
-}
-
-struct URLSessionHttpClientEngine: HttpClientEngine {
-    func send(request: URLRequest) async throws -> (Data, URLResponse) {
-        try await URLSession.shared.data(for: request)
-    }
-}
-
 final public class AuthenticationClientAuthLib: AuthenticationClient {
     public init() {}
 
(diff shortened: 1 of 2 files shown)


Projects/App/Sources/Service/OctopusClientsImplementation/AuthenticationClientAuthLib.swift
```diff
@@ -6,35 +6,41 @@ import Foundation
...
+    func send(request: URLRequest) async throws -> (Data, URLResponse) {
+        try await URLSession.shared.data(for: request)
+    }
+}
+
 final public class AuthenticationClientAuthLib: AuthenticationClient {
     public init() {}
 
     @MainActor
-    private var networkAuthRepository: NetworkAuthRepository = {
```
sladan-hedvig, line 13: Just replacing this networkAuthRepository with
```
private lazy var networkAuthRepository: NetworkAuthRepository = {[weak self] in
        return NetworkAuthRepository(
            environment: Environment.current.authEnvironment,
            additionalHttpHeadersProvider: {
                var headers = self?.getHeaders() ?? [:]
                return headers
            },
            httpClientEngine: nil
        )
    }()
    
    private func getHeaders() -> [String: String] {
        var headers = [String: String]()
        let semaphore = DispatchSemaphore(value: 0)
        Task {
            headers = await ApolloClient.headers()
            semaphore.signal()
        }
        semaphore.wait()
        return headers
    }
```
with do the work.

Instead of creating new instance of NetworkAuthRepository every time, we want to make headers dynamic only.

Tuist/Config.swift
```diff
@@ -2,7 +2,7 @@ import ProjectDescription
 
 let config = Config(
     compatibleXcodeVersions: .list([
-        .upToNextMajor(.init(16, 0, 0))
```
sladan-hedvig, line 5: Can we also keep 
`.upToNextMajor(.init(16, 0, 0))
` so we can support both until we move to Xcode 26. ?

sladan-hedvig (changes requested):

Tuist/Config.swift
sladan-hedvig: You can do it by setting:
.upToNextMajor(.init(16, 0, 0)),
.upToNextMajor(.init(26, 0, 0))

sladan-hedvig (approved):

PR #1928, opened 2025-07-02
GEN-3910: iOS26 Help center search bar fix

juliaendre:
## [GEN-3910]

- **Problem:** Search bar is moved to bottom in iOS 26 and hides support view.
- **Solution:** Added more padding for iOS26 to not be hidden by search bar
- My first approach was to make the search bar stick to the top but it seems like we would have to create our own custom search bar in that case. Thought this would be easier for now just to make sure that the support view is not hidden and we can update this later on if we feel like it. I also know that we have talked about changing/removing the contact view here so doesn't feel worth spending that much time on it.

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


juliaendre:
True, will take this into consideration!

juliaendre:
@sladan-hedvig updated!

sladan-hedvig (approved):

PR #1935, opened 2025-07-07
Migrate from authlib to umbrella

StylianosGakis:
https://github.com/HedvigInsurance/umbrella

Please check that everything works as intended, and that the changes I made inside post-build-action.sh make sense too, as I am not sure how I would properly test that myself


sladan-hedvig (approved):
AMAZING 👍

PR #1995, opened 2025-08-19
Merge edit tier screens

juliaendre:
## [APP-XXX]

- Merge edit tier screens since they have basically have the same content

## Checklist

- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig (commented):
Looks good, but shouldn't we say "From " when seleceting tier?

juliaendre:
@sladan-hedvig yes that was updated in the moving flow PR so thought I could skip it here, but added it here as well now

sladan-hedvig (approved):
🥳

PR #2025, opened 2025-09-15
GEN-4170: Addon Graphql API Integration

juliaendre:
## [GEN-4170]

- Graphql api integration
- Added displayNameLong as costBreakdown items in summary screen with prices

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Addons/Sources/Views/ChangeAddonSummaryScreen.swift
+++ b/Projects/Addons/Sources/Views/ChangeAddonSummaryScreen.swift
@@ -52,11 +52,8 @@ extension ChangeAddonViewModel {
             isAddon: true,
             summaryDataProvider: DirectQuoteSummaryDataProvider(
                 intentCost: .init(
-                    gross: self.addonOffer?.currentAddon?.itemCost.premium.gross,
-                    net: getTotalPrice(
-                        currentPrice: addonOffer?.currentAddon?.itemCost.premium.net,
-                        newPrice: selectedQuote?.itemCost.premium.net
-                    )
+                    gross: nil,
+                    net: getTotalPrice()
                 )
             )
         ) { [weak self, weak changeAddonNavigationVm] in
--- a/Projects/Addons/Sources/Views/ChangeAddonSummaryScreen.swift
+++ b/Projects/Addons/Sources/Views/ChangeAddonSummaryScreen.swift
@@ -27,13 +27,9 @@ extension ChangeAddonViewModel {
             exposureName: L10n.addonFlowSummaryActiveFrom(
                 addonOffer?.activationDate?.displayDateDDMMMYYYYFormat ?? ""
             ),
-            //            premium: .init(
-            //                gross: addonOffer?.currentAddon?.price,
-            //                net: selectedQuote?.
-            //            ),
             premium: .init(
-                gross: .sek(0),
-                net: .sek(0)
+                gross: addonOffer?.currentAddon?.price.gross,
+                net: addonOffer?.currentAddon?.price.net
             ),
             documentSection: .init(
                 documents: selectedQuote?.documents ?? [],
@@ -48,7 +44,7 @@ extension ChangeAddonViewModel {
             insuranceLimits: [],
             typeOfContract: nil,
             isAddon: true,
-            priceBreakdownItems: []
+            priceBreakdownItems: getBreakdownDisplayItems()
         )
 
         let vm = QuoteSummaryViewModel(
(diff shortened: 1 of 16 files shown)


Projects/hCoreUI/Sources/Views/QuoteSummaryScreen/QuoteSummaryViewModel.swift
```diff
@@ -170,7 +170,7 @@ public class QuoteSummaryViewModel: ObservableObject, Identifiable {
 public struct QuoteDisplayItem: Identifiable, Equatable, Sendable {
     public let id: String?
     let displayTitle: String
-    let displayValue: String
+    let displayValue: String?
     let displayValueOld: String?
```
sladan-hedvig, line 174: I think we can remove `displayValueOld` and use some flag to tell us if we should cross  displayTitle and displayValue. 
It was used only by addons and we don't need it anymore since we don't want to show crossed `45 days` and `60 days `next to it
We will have it in the breakdown just above card's total breakdown.

Projects/Addons/Sources/Views/ChangeAddonSummaryScreen.swift
```diff
@@ -55,11 +55,10 @@ extension ChangeAddonViewModel {
             isAddon: true,
             summaryDataProvider: DirectQuoteSummaryDataProvider(
                 intentCost: .init(
```
sladan-hedvig, line 57: gross should be always null here since we don't want to show price before in total cost.
and net should be dynamic:
`func getTotalPrice(currentPrice: MonetaryAmount?, newPrice: MonetaryAmount?) -> MonetaryAmount {
`can be refactored into 
```
func getTotalPrice() -> MonetaryAmount {
        guard let selectedQuoteNet = selectedQuote?.price.net else {
            return .init(amount: 0, currency: "SEK")
        }
        
        if let currentAddonNet = addonOffer?.currentAddon?.price.net {
            let amount = selectedQuoteNet.floatAmount - currentAddonNet.floatAmount
            return .init(amount: amount, currency: selectedQuoteNet.currency)
        }
        return selectedQuoteNet
}
```
Or something simular

Projects/Addons/Sources/Views/ChangeAddonSummaryScreen.swift
```diff
@@ -28,8 +28,8 @@ extension ChangeAddonViewModel {
                 addonOffer?.activationDate?.displayDateDDMMMYYYYFormat ?? ""
             ),
             premium: .init(
```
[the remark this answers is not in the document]
juliaendre: Changed it. Please have a look!

Projects/App/Sources/Service/OctopusClientsImplementation/AddonsClientOctopus.swift
```diff
@@ -22,17 +22,22 @@ class AddonsClientOctopus: AddonsClient {
...
                         .init(fragment: $0.fragments.upsellTravelAddonDisplayItemFragment)
                     },
-                    price: .init(fragment: currentAddon.premium.fragments.moneyFragment),
+                    itemCost: .init(
+                        premium: .init(gross: nil, net: .init(fragment: currentAddon.netPremium.fragments.moneyFragment)),
+                        discounts: []
+                    ),
                     addonVariant: nil,
-                    documents: []
+                    documents: [],
```
sladan-hedvig, line 40: I am just curious, why you keep adding this , on the last argument?

Projects/hCoreUI/Sources/Views/QuoteSummaryScreen/QuoteSummaryViewModel.swift
```diff
@@ -171,17 +171,17 @@ public struct QuoteDisplayItem: Identifiable, Equatable, Sendable {
     public let id: String?
```
sladan-hedvig, line 171: Not sure why this was optional from the start. Can we make it not optional and just set
`self.id = id ?? UUID().uuidString`

sladan-hedvig (approved):
Nothing major, just that id

Projects/App/Sources/Service/OctopusClientsImplementation/AddonsClientOctopus.swift
juliaendre: What do you think?

juliaendre: I don't care really so we can remove them

PR #2044, opened 2025-09-26
Updates for disabling liquid glass on Xcode 26

sladan-hedvig:
## [APP-XXX]

- Disable liquid glass on the iOS 26
- Do some tweeks if liquid glass is enabled (still in experimental phase)
- App should look the same on iOS 18 and iOS 26
After merging this we can safely move to the Xcode 26
## Checklist

- [ ] Landscape verification
- [ ] iPad verification


Projects/hCoreUI/Sources/PresentationStyle+Detented.swift
```diff
@@ -108,7 +108,9 @@ class DetentTransitioningDelegate: NSObject, UIViewControllerTransitioningDelega
                     presenting: presenting,
                     useBlur: options.contains(.useBlur)
                 )
-                presentationController.preferredCornerRadius = .cornerRadiusXL
+                if !isLiquidGlassEnabled {
+                    presentationController.preferredCornerRadius = .cornerRadiusXL
```
sladan-hedvig, line 112: If we specify corner radius, it will not play nicely with glass effects. Set this only when we aren't using glass

juliaendre:
simulator_screenshot_EAEEF6C7-7B91-4B4F-B7AA-95D69FC99593
This looks horrible on glass look 🫠 but I guess we'll fix that later on

juliaendre (approved):

PR #2057, opened 2025-10-03
GEN-4316: Only show “Hedvig Service Team” if first message after automation

juliaendre:
## [GEN-4316]

- This regards [this slack conversation] (https://hedviginsurance.slack.com/archives/C08C1DG4TFD/p1759482879804499)
- Show only "Hedvig Service Team" if it's the first Hedvig sent message after automation

## Checklist

- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Chat/Sources/Views/ChatScreenViewModel.swift
+++ b/Projects/Chat/Sources/Views/ChatScreenViewModel.swift
@@ -31,6 +31,20 @@ public class ChatMessageViewModel: ObservableObject {
     @Published var messages: [Message] = []
     @Published var isFetchingPreviousMessages = false
 
+    func showHedvigTimeStamp(message: Message) -> Bool {
+        let hasAutomatedMessages = messages.first(where: { $0.sender == .automation }) != nil
+        return message.sender == .hedvig && hasAutomatedMessages && isFirstHedvigMessage(currentMessageId: message.id)
+    }
+
+    private func isFirstHedvigMessage(currentMessageId: String) -> Bool {
+        guard let currentMessageIndex = messages.firstIndex(where: { $0.id == currentMessageId }),
+            currentMessageIndex != 0
+        else { return false }
+        let previousMessage = messages[currentMessageIndex + 1]
+
+        return previousMessage.sender != .hedvig
+    }
+
     public init(
         chatService: ChatServiceProtocol
     ) {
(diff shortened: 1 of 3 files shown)


Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -31,6 +31,20 @@ public class ChatMessageViewModel: ObservableObject {
...
+    func showHedvigTimeStamp(message: Message) -> Bool {
+        let hasAutomatedMessages = messages.first(where: { $0.sender == .automation }) != nil
+        return message.sender == .hedvig && hasAutomatedMessages && isFirstHedvigMessage(currentMessageId: message.id)
+    }
+
+    private func isFirstHedvigMessage(currentMessageId: String) -> Bool {
+        guard let currentMessageIndex = messages.firstIndex(where: { $0.id == currentMessageId }),
+            currentMessageIndex != 0
+        else { return false }
+        let previousMessage = messages[currentMessageIndex + 1]
```
sladan-hedvig, line 43: Be careful when fetching like this, it can cause crash if there is no element in the array at that index.

Projects/Chat/Sources/Views/ChatScreenViewModel.swift
```diff
@@ -31,6 +31,20 @@ public class ChatMessageViewModel: ObservableObject {
     @Published var messages: [Message] = []
     @Published var isFetchingPreviousMessages = false
 
+    func showHedvigTimeStamp(message: Message) -> Bool {
+        let hasAutomatedMessages = messages.first(where: { $0.sender == .automation }) != nil
```
sladan-hedvig, line 35: This on the demand seems like a lot of computing.
Can we have it like 2 properties:
```
    private var hasAutomation: Bool = false
    private(set) var firstHedvigMessageAfterAutomation: Message?
```
where we will compute when we need.

Inside 
`func fetchPreviousMessages(retry withAutomaticRetry: Bool = true) async {
`
we could set automation 
```
...
let newMessages = chatData.messages.filterNotAddedIn(list: addedMessagesIds)
                if newMessages.contains(where: { $0.sender == .automation }) {
                    hasAutomation = true
                }
                withAnimation {
...
```
same for fetchMessages
```
                let newMessages = chatData.messages.filterNotAddedIn(list: addedMessagesIds)
                if newMessages.contains(where: { $0.sender == .automation }) {
                    hasAutomation = true
                }
```

and update messages with
```
    @Published var messages: [Message] = [] {
        didSet{
            if hasAutomation {
                firstHedvigMessageAfterAutomation = messages.last(where: {$0.sender == .hedvig})
            }
        }
    }
```
This will reduce computing and do calculation when it is needed

sladan-hedvig (changes requested):

PR #2363, opened 2026-04-13
Eng: Kmp provided help center feature. Experimental

StylianosGakis:
[DO NOT MERGE]

Consumes the compose UI code to have the entire help-center be served from common code, from the UI, all the way up to the networking.

Can be used just as a sample of how this _could_ be done. For this to work there were a lot of things simply stubbed out to make the app simply work, like not hooking up to the real feature flags, not passing the real constants to the backend like model, manufacturer and so on.

The real auth token is used however, simply delegating to the existing iOS implementation for all of that.

`KMPHelpCenterView` was claude-generated, I didn't check for correctness. It's there to show that some part of the app can be shared while still being able to delegate to other parts of the app without any issues.


StylianosGakis:
Superceded by https://github.com/HedvigInsurance/ugglan/pull/2418

PR #2414, opened 2026-05-06
feat: add missing payments info card and rename state view environment keys (GEN-5216)

sladan-hedvig:
## Summary
- Add info card with cancellation warning on missed payment screen when payment status is `contactUs`
- Rename state view environment keys (`hSuccessBottomAttachedView` → `hStateViewBottomAttachedView`, `hCustomSuccessView` → `hStateViewCustomSuccessView`) and move definitions to `StateView.swift`
- Add new `hStateViewContentBottomAttachedView` environment key for attaching content below state view body

## Test plan
- [ ] Verify missed payment screen shows info card when status is `contactUs`
- [ ] Verify missed payment screen hides info card for other statuses
- [ ] Verify success modal shows info card when status is `contactUs`
- [ ] Verify existing processing/success screens still work with renamed environment keys

### DIFF
--- a/Projects/App/Sources/Navigation/LoggedInNavigation.swift
+++ b/Projects/App/Sources/Navigation/LoggedInNavigation.swift
@@ -888,12 +888,17 @@ class LoggedInNavigationViewModel: ObservableObject {
     @Published var askForPushNotification = false
     @Published var isReviewContactInfoPresented = false
     @Published var missedPaymentData: MissedPaymentData?
+    @Published var hasMissedPayment = false
 
     private var cancellables = Set<AnyCancellable>()
     weak var tabBar: UITabBarController? {
         didSet {
-            guard #available(iOS 18, *), UIDevice.current.userInterfaceIdiom == .pad else { return }
-            tabBar?.traitOverrides.horizontalSizeClass = .compact
+            if #available(iOS 18, *), UIDevice.current.userInterfaceIdiom == .pad {
+                tabBar?.traitOverrides.horizontalSizeClass = .compact
+            }
+            Task { @MainActor [weak self] in
+                self?.updatePaymentsBadge()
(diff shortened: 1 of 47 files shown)


Projects/App/Sources/Navigation/LoggedInNavigation.swift
```diff
@@ -1062,6 +1062,7 @@ class LoggedInNavigationViewModel: ObservableObject {
     @objc func petChipIdAdded() {
         let homeStore: HomeStore = globalPresentableStoreContainer.get()
         homeStore.send(.fetchMemberState)
+        homeStore.send(.fetchMissedCharge)
```
sladan-hedvig: Added by accident, removed in the following commit

Projects/Home/Sources/Screens/HomeScreen.swift
```diff
@@ -154,6 +155,9 @@ class HomeVM: ObservableObject {
         store.send(.fetchImportantMessages)
         store.send(.fetchQuickActions)
         store.send(.fetchChatNotifications)
+        if store.state.hasMissedCharge {
+            store.send(.fetchMissedCharge)
+        }
```
StylianosGakis, line 160: Is this `fetchHomeState` the part which will happen on a loop? So that we continuously fetch this only if it was `hasMissedCharge = true` before?

StylianosGakis (approved):

Projects/Home/Sources/Screens/HomeScreen.swift
sladan-hedvig: This will only happen when we open home screen, but I set hasMissedCharge to false after we successfully pay for missed charge, so no loop from our side - feels like overkill to do it every 15s.

PR #2459, opened 2026-06-05
eng: remove campaign store (GEN-5368)

sladan-hedvig:
## Summary
Another step in moving the app off the legacy `PresentableStore` pattern. The payments discounts screen now drives its state from a plain SwiftUI ViewModel instead of `CampaignStore`, in line with the architecture direction we've been heading toward across the codebase.

Nothing user-facing changes — the screen still fetches the same data from the same service and renders the same UI for the same states (loading, success, error). The underlying machinery is just simpler now: less indirection, no global store, and the screen's lifecycle drives the fetch directly.

Existing store tests were rewritten against the ViewModel and continue to cover the same success/failure paths.

## Test plan
- [ ] Open the Payments tab → tap "Discounts" → confirms list renders and the referral info button works.
- [ ] Force an error (e.g. offline) → confirms the retry button re-fetches.
- [ ] `Campaign` unit tests pass.


sladan-hedvig:
Create another PR with additional features

PR #2512, opened 2026-07-08
Fix iOS issues from travel submission flow test session (RND-2030)

hugokallstrom:
## Why

The [Deterministic travel submission flow 2 test session](https://app.notion.com/p/hedviginsurance/Deterministic-travel-submission-flow-2-397c3f71cb0a8060bafec55ebda17c8a) reported iOS issues (RND-2030). This PR fixes the two claim-chat UI ones:

1. Form text fields (e.g. bike theft lock question) use a small single-line field that breaks on answers longer than a screen width — resolution: use the regular text input.
2. Free-text answers render as a full-width pill, which looks broken now that the 50 character minimum is gone — resolution: remove min width requirements.

The third finding from the session (select bottom sheet height / native iOS 26 sheet) is split into its own PR since it touches the shared detent presentation code.

## What

- `.text` form fields in the claim chat now use the multi-line `hTextView` free-text input, with min/max characters wired from the backend field, and section padding restored inside the field.
- `hTextView`'s character counter is hidden when no max length is provided.
- The submitted free-text pill hugs its content instead of stretching full width, and text answers align trailing like other member responses.

Verified on an iOS 26.5 simulator by driving the flow end-to-end (short 2-char answer → hugging pill; 120-char answer → wraps over 3 lines in editor, field, and result bubble).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2515, opened 2026-07-09
Use native sheet presentation for detents on iOS 26 (RND-2030)

hugokallstrom:
## Why

From the [travel submission flow test session](https://app.notion.com/p/hedviginsurance/Deterministic-travel-submission-flow-2-397c3f71cb0a8060bafec55ebda17c8a): the select bottom sheet opens at the wrong height on iOS 26, and the ticket asks for the native iOS 26 sheet instead of our custom one.

Today every `.detent(...)` presentation uses `modalPresentationStyle = .custom` with our own presentation controller, built on private UIKit API (`_UISheetDetent`, `_setIndexOfCurrentDetent`). It also presents at a zero-height detent and swaps in the real detents 100 ms later — which is why sheets open at the wrong height and jump.

Split out of #2512 so that PR stays a simple UI review; this one changes shared presentation code and deserves its own look.

## What

One block in `DetentRouter.handle(isPresent:)`, gated behind `if #available(iOS 26.0, *)`:

- Present with the standard `.pageSheet`, so UIKit provides the native iOS 26 sheet (grabber, corner treatment, dimming) — no private API involved.
- Map our `Detent` styles through the existing public `Detent.set` helper, applied **before** presentation instead of 100 ms after.
- Lay out the content once pre-presentation so the `.height` detent resolves from the real scroll view content size instead of 0.
- iOS 16–18 keep the existing custom path, byte-for-byte unchanged (the delegate is now also kept alive through the presentation animation, which was previously only luck).

Verified by driving the claim flow on an iOS 26.5 simulator: honesty pledge and lock-type select sheets both open natively at exactly content height, no flash or jump.

### DIFF
--- a/Projects/Chat/Sources/Views/ChatScreen.swift
+++ b/Projects/Chat/Sources/Views/ChatScreen.swift
@@ -287,7 +287,7 @@ class ChatScrollViewDelegate: NSObject, UIScrollViewDelegate, ObservableObject {
             if let navigation = vc.navigationController {
                 return findProverVC(from: navigation)
             } else {
-                if vc.presentationController is BlurredSheetPresentationController {
+                if vc.presentationController is UISheetPresentationController {
                     return vc
                 } else if let superviewVc = vc.view.superview?.viewController {
                     return findProverVC(from: superviewVc)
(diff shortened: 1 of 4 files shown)


StylianosGakis (approved):
Approve to unblock but would be nice with another pair of eyes later.

sladan-hedvig:
@hugokallstrom can we close this PR and use this [one](https://github.com/HedvigInsurance/ugglan/pull/2525) instead? It should resolve the issue.

sladan-hedvig:
Also, when we move to iOS 17, we will revisit this and make it properly.

PR #2516, opened 2026-07-10
Support the claim chat Information step (RND-2030)

hugokallstrom:
## Why

The agent-driven claim flow in claims-service can return a `ClaimIntentStepContentInformation` step, but iOS had no fragment selection or model case for it, so the step mapping failed and the member dead-ended. This exact gap broke Android fire-claim submissions on staging, so this mirrors the Android fix (HedvigInsurance/android#3017).

## What

- Add `ClaimIntentStepContentInformation` (`notice`, `severity`, `buttonTitle`) to `ClaimIntentStepContentFragment` and a new `ClaimIntentSubmitInformation` mutation.
- Add a `.information` case to `ClaimIntentStepContent` with a model struct; unknown severities degrade to `INFO` instead of failing the flow.
- Add `claimIntentSubmitInformation(stepId:)` to `ClaimIntentClient` (protocol, Octopus, and demo implementations) plus a `SubmitClaimInformationStep` handler.
- Render the notice as an `InfoCard` (red for `CRITICAL`, blue for `INFO`) with a server-driven acknowledge button that submits and advances the flow; critical notices are announced to VoiceOver.
- Tests for the step handler and the fragment→model mapping (including the unknown-severity degrade).

### DIFF
--- a/Projects/App/Sources/Service/OctopusClientsImplementation/SubmitClaim/ClaimIntentClientOctopus.swift
+++ b/Projects/App/Sources/Service/OctopusClientsImplementation/SubmitClaim/ClaimIntentClientOctopus.swift
@@ -162,6 +162,23 @@ class ClaimIntentClientOctopus: ClaimIntentClient {
         }
     }
 
+    func claimIntentSubmitInformation(stepId: String) async throws -> ClaimIntentType? {
+        let input = OctopusGraphQL.ClaimIntentSubmitInformationInput(stepId: stepId)
+        let mutation = OctopusGraphQL.ClaimIntentSubmitInformationMutation(input: input)
+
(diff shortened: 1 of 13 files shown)


StylianosGakis (approved):
Approve to unblock but would be nice with another pair of eyes later.

sladan-hedvig:
@hugokallstrom this step has been already implemented and now we have conflicts. I am not sure if we want to have it as existing implementation or as it is done here.

hugokallstrom:
> @hugokallstrom this step has been already implemented and now we have conflicts. I am not sure if we want to have it as existing implementation or as it is done here.

Lets use the existing, ill close this.

PR #2528, opened 2026-07-21
eng: Added dev settings

sladan-hedvig:
## [APP-XXX]

- Added dev settings
- For now only to disable honesty pledge/animations when submitting claim 
- Built upon resume claim since less friction and conflicts 

## Blockers

- Blocker

## Checklist

- [ ] Unit tests pass
- [ ] Landscape verification
- [ ] iPad verification


Projects/Profile/Sources/Views/Screens/Profile/Profile.swift
```diff
@@ -22,8 +23,14 @@ public struct ProfileView: View {
                 }
                 ProfileRow(row: .claimHistory)
                 ProfileRow(row: .information)
-                ProfileRow(row: .settings)
-                    .hWithoutDivider
+                if Environment.current != .production {
```
StylianosGakis, line 26: is "staging" also != production? Or is this really just for local Xcode builds?

StylianosGakis (approved):

Projects/Profile/Sources/Views/Screens/Profile/Profile.swift
sladan-hedvig: Yeah! Hedvig will never have this option

PR #2580, opened 2026-08-18
feat: Home Screen 6 — ongoing quotes

victor-reyes-by:
## [RND-2069]

- "Your quotes" section on the home sheet: unfinished web shop sessions, resumed in the device browser via `resumeUrl`
- Gated in the data layer by the inverted Unleash kill switch `disable_resuming_ongoing_shop_sessions`, so the section self-hides when the flag is off
- `ongoingQuotes` is `@Transient` — a quote can be signed or expire on the web while the app is closed; the backend already excludes signed and expired sessions, so there is no client-side filtering

## Blockers

- None

## Checklist

- [ ] Unit tests pass
- [ ] Landscape verification
- [ ] iPad verification


Projects/Home/Sources/Screens/Components/HomeOngoingQuotesSection.swift
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private func resume() {
+        log.addUserAction(
+            type: .click,
+            name: "home ongoing quote",
+            attributes: ["quoteId": quote.id]
+        )
+        Task {
+            await Dependencies.urlOpener.open(quote.resumeUrl)
```
sladan-hedvig, line 91: Can you also add loader for the view while we await for opening URL

sladan-hedvig (approved):

sladan-hedvig (approved):

PR #2582, opened 2026-08-20
fix: Onboarding test session fixes (GEN-5511)

sladan-hedvig:
## Description

Fixes for the iOS items from the onboarding test session ([Notion page](https://app.notion.com/p/hedviginsurance/Onboarding-3c2c3f71cb0a80ce97cddd8937206d9f)):

- **Showing rate** — skip asking for App Store review while onboarding is enabled
- **Skip analytics screen until legal is done** — added analytics feature flag
- **Data consent Allow/Deny** — buttons are no longer disabled after selection; the selected choice is highlighted instead
- **Move theme step after "bank"** — theme step now comes after the connect payment step
- **Animate checkmark after adding co-insured info / chip ID** — added delay for added missing info + transitions when returning to the onboarding flow
- **"Do it later" button when info is already added** — only the continue button is shown on the missing info screen when everything is filled in
- **Chip ID screen font size / updated Figma design** — removed ContractInformation size option
- **Number pad for screen readers** — number pad is hidden when the keyboard is presented

### DIFF
--- a/Projects/hCore/Sources/AskForRating.swift
+++ b/Projects/hCore/Sources/AskForRating.swift
@@ -11,6 +11,7 @@ public struct AskForRating {
         UserDefaults.standard.set(numberOfSessions, forKey: userDefaultsKey)
     }
 
+    @MainActor
     public func askAccordingToTheNumberOfSessions() {
         guard !UserDefaults.standard.bool(forKey: userDefaultsCompletedKey) else { return }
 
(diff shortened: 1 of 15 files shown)


Projects/hCore/Sources/AskForRating.swift
```diff
@@ -21,7 +22,11 @@ public struct AskForRating {
         }
     }
 
+    @MainActor
     public func askForReview() {
+        if Dependencies.featureFlags().isOnboardingEnabled && !OnboardingPresentationState.hasSeenOnboarding {
+            return
+        }
```
victor-reyes-by, line 29: why not `guard` here?😄

victor-reyes-by (approved):
🚀

Projects/hCore/Sources/AskForRating.swift
sladan-hedvig: This is fine also no?

PR #2592, opened 2026-08-25
Payment protection: 90 day qualification info on change amount summary

sladan-hedvig:
[GEN-5624](https://app.notion.com/p/hedviginsurance/App-If-increased-amount-in-change-amount-flow-Information-box-about-90-day-qualification-period-3c2c3f71cb0a80729756c7847c13419a)

When a member **increases** their payment protection amount, the change amount summary screen now shows an info box about the 90 day qualification period before the higher amount takes effect.

## Changes

- `ChangeTierViewModel.summaryNoticeInfo` returns `CHANGE_TIER_PAYMENT_PROTECTION_INFO` when the contract is payment protection and the selected tier level is higher than the current one — `nil` otherwise, so lowering or keeping the amount shows nothing.
- `ChangeTierSummaryScreen` passes it into `QuoteSummary.noticeInfo`, which renders as an `InfoCard` above the price section (existing hook, already used by MoveFlow and Addons).
- Tests cover increase, no increase, and non–payment protection contracts.

## Note

`CHANGE_TIER_PAYMENT_PROTECTION_INFO` is still empty in `sv-SE`. Payment protection is SE-only, so the translation needs to land before this reaches members or the info box renders empty.

### DIFF
--- a/Projects/ChangeTier/Sources/Views/ChangeTierSummaryScreen.swift
+++ b/Projects/ChangeTier/Sources/Views/ChangeTierSummaryScreen.swift
@@ -51,6 +51,7 @@ extension ChangeTierViewModel {
         return QuoteSummary(
             contracts: contracts,
             activationDate: self.activationDate,
+            noticeInfo: summaryNoticeInfo,
             totalPrice: .comparison(old: totalPremium.gross, new: totalPremium.net)
         )
     }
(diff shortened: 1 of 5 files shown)


victor-reyes-by (approved):
Nice! 
Isn't there too much typeOfContract/PPI logic there in `ChangeTierViewModel`, though? Feels like it should be more generic? Idk, though, just thinking aloud 🫪🤓

sladan-hedvig:
Will control it from the BE, so closing it

PR #2599, opened 2026-08-28
feat: Home Screen — fade sheet content under the pinned header

sladan-hedvig:
## [RND-XXXX]

Replaces the home sheet's hard top clip with a two-stage mask, so content dissolves as it scrolls under the pinned header instead of cutting off mid-glyph.

- Swap `TopClipShape` for two `.mask` passes with the sheet surface sandwiched between them: the first fades the content alone over 4pt, the second hard-cuts content *and* surface at the header's bottom edge so the sheet keeps a crisp top edge
- Carry the old `bottomExtension` behaviour across to the masks via negative bottom padding, so the 500pt overscroll surface still covers rubber-banding at the bottom rather than exposing the hero
- Pad the sheet content down by the fade height, keeping the first card clear of the fade band at rest
- Tighten the grabber's bottom spacing in `HomeSheetContainer` (`.padding10` top, `.padding6` bottom)

## Blockers

- None

## Checklist

- [ ] Unit tests pass
- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Home/Sources/Screens/ActiveHomeView.swift
+++ b/Projects/Home/Sources/Screens/ActiveHomeView.swift
@@ -26,6 +26,9 @@ struct ActiveHomeView: View {
     /// White below the sheet's end so rubber-banding never exposes the hero through the bottom.
     private let bottomOverscrollExtension: CGFloat = 500
 
+    /// How far the content dissolves before it slips under the pinned header.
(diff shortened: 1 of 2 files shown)


victor-reyes-by (approved):
Looks great!

PR #2605, opened 2026-08-31
fix: set proper place for file picker

sladan-hedvig:
- alert for selecting source of files was showing on whole elements instead of just a button where it is called from

## Blockers

- Blocker

## Checklist

- [ ] Unit tests pass
- [ ] Landscape verification
- [ ] iPad verification


victor-reyes-by (approved):
Nice catch!

PR #2612, opened 2026-09-02
chore: round all buttons

sladan-hedvig:
Filled buttons are now always pill-shaped, instead of being rounded per-size with an opt-in override at a few call sites.

- `ButtonFilledStyle` uses `.cornerRadiusRounded` unconditionally; `hButtonSize.cornerRadius` (`cornerRadiusS`/`M`/`L`) no longer feeds into it
- Removed the `hCustomButtonCornerRadius` environment value and its `View` modifier, plus the two call sites that opted in (`HomeActionChips`, `OnboardingNavigation`) — they now get the rounded shape by default
- `cornerRadiusRounded` changed from `999` to `36`. With the capsule hack gone it only needs to exceed half the tallest button height, and a real value keeps `RoundedRectangle` from over-clipping

## Checklist

- [ ] Unit tests pass
- [ ] Landscape verification
- [ ] iPad verification

### DIFF
--- a/Projects/Home/Sources/Screens/Components/HomeActionChips.swift
+++ b/Projects/Home/Sources/Screens/Components/HomeActionChips.swift
@@ -73,7 +73,6 @@ struct HomeActionChips: View {
                     .padding(.top, .padding8)
                     .padding(.bottom, .padding6)
             }
-            .hCustomButtonCornerRadius(.cornerRadiusRounded)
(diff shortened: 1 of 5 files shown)


victor-reyes-by:
> * With the capsule hack gone

👀

victor-reyes-by (approved):
Nice! 🤩

PR #2615, opened 2026-09-03
chore: suppress duplicate Datadog logs within 0.5s

sladan-hedvig:
## What

`DatadogLogger` now drops an identical message if the same one was already logged less than half a second ago. A repeating log — a view redraw, a retry loop — no longer floods Datadog with the same line.

All six level methods (`debug`/`info`/`notice`/`warn`/`error`/`critical`) guard on a new `shouldLog(_:level:)`. `addUserAction` and `addError` are untouched, since those are RUM actions/errors rather than log messages.

## Notes on the implementation

- **Keyed on level + message**, not message alone. Keying on the message alone would let an `info("Fetching")` swallow a following `error("Fetching")` — the kind of suppression that hides real problems. `error:` and `attributes:` are not part of the key, so message identity alone decides.
- **`ProcessInfo.processInfo.systemUptime`** rather than `Date()`, so a wall-clock change can't produce a negative delta and wrongly let a duplicate through.
- **`NSLock` around the dictionary.** `Logging` isn't `@MainActor`, so logs arrive on arbitrary threads and unsynchronized writes to a `Dictionary` can corrupt its buffer mid-resize, not merely miss a dedup.
- **The emission stays outside the lock.** `datadogLogger.x(...)` is called after the unlock, so the actual log write isn't serialized and there's no deadlock risk from the Datadog SDK taking its own locks.
- **Cache capped at 200 entries**, pruning expired ones on overflow. Log messages routinely embed dynamic data (IDs, URLs, error text), so a high-cardinality stream would otherwise grow the cache forever.

## Alternatives considered

- An `actor` for the cache — rejected because `Logging`'s methods are synchronous, so the emission would have to move inside a `Task`, delaying and reordering logs. A lot of logging happens right before a crash, which is exactly when that loss matters.
- A `Set` plus a delayed `Task` to evict each entry — self-pruning, but it still needs the lock (the eviction is a second writer), costs a timer per log line, and makes the window imprecise under load.

One follow-up left on the table: `NSLock` has no priority donation, so a low-QoS thread holding it while descheduled could stall a waiting main thread. With a critical section this small the odds are remote; `OSAllocatedUnfairLock` is a drop-in swap if it ever shows up in a trace.

## Testing

Not covered by tests, and not compiled locally — SwiftLint isn't installed on this machine and the build was left to Xcode/CI. Behaviour is unverified beyond code reading.

### DIFF
--- a/Projects/App/Sources/DatadogLogger.swift
+++ b/Projects/App/Sources/DatadogLogger.swift
@@ -3,20 +3,51 @@ import DatadogInternal
 import DatadogLogs
 import DatadogRUM
 import Environment
+import Foundation
 import Logger
 import SwiftUI
 
 class DatadogLogger: Logging {
+    /// Identical messages logged within this window are dropped, so a repeating log
+    /// (a redraw, a retry loop) does not flood Datadog.
+    private static let duplicateWindow: TimeInterval = 0.5
+    /// Upper bound on tracked messages, so a stream of unique messages cannot grow the cache forever.
+    private static let maxTrackedMessages = 200
+
+    private struct LogKey: Hashable {
+        let level: String
+        let message: String
+    }
+
     private let datadogLogger: LoggerProtocol
+    private let lock = NSLock()
+    private var lastLoggedAt: [LogKey: TimeInterval] = [:]
 
     init(datadogLogger: LoggerProtocol) {
         self.datadogLogger = datadogLogger
     }
 
-    func debug(_ message: String, error: Error? = nil, attributes: [AttributeKey: AttributeValue]? = nil) {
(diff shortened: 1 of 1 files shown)


victor-reyes-by (approved):
Nice, but feels like we're hiding problem, instead of solving it 👾

sladan-hedvig:
> Nice, but feels like we're hiding problem, instead of solving it 👾

It was mostly for "Access token refresh missing token"
 and 
"Refresh token expired at 2026-09-02 11:33:28 +0000 forcing logout"

sladan-hedvig:
@victor-reyes-by has a point here! Will close

PR #2616, opened 2026-09-03
fix: iOS 16 stack-overflow crash on logout teardown

sladan-hedvig:
## The crash

Recurring on **iOS 16 only**, when the user **logs out** (or is force–logged out).

The crash log is all system frames repeating the same release/deinit call over and over — a classic **stack overflow while tearing down the view tree**.

## Why it happens

On logout we switch the root screen from the logged-in view to the login view inside `withAnimation`. The animation keeps the whole logged-in view alive and deallocates it all at once. `ActiveHomeView` is very deeply nested, so that one big teardown runs off the stack.

iOS 17+ handles the same teardown fine — only iOS 16 crashes.

## The fix

1. **`MainNavigation.swift`** — on iOS 16 only, do the logout screen switch **without** animation, so the tree isn't all torn down in one pass. iOS 17+ keeps the animation. The launch/login screen still covers the transition.
2. **`ActiveHomeView.swift`** — split the screen into smaller `View` structs (`HeroBackgroundView`, `HomeSheetContent`, `TopFadeMask`) so the view tree isn't one giant nested value. No visual/behavior change.

## Testing

- Builds clean (no Xcode issues).
- ⚠️ Please verify on an **iOS 16 device/simulator**: log in → open Home → log out and force-logout a few times. I couldn't reproduce iOS 16 teardown locally.

It mist be related to the specific 16.7.16 OS version....

### DIFF
--- a/Projects/Home/Sources/Screens/ActiveHomeView.swift
+++ b/Projects/Home/Sources/Screens/ActiveHomeView.swift
@@ -9,8 +9,6 @@ import hCoreUI
 
 struct ActiveHomeView: View {
     @AppObservedObject private var homeStore: HomeStore
-    @AppObservedObject private var claimsStore: ClaimsStore
-    @AppObservedObject private var crossSellStore: CrossSellStore
     @StateObject private var bottomVm = HomeBottomScrollViewModel()
     @State private var greetingHeight: CGFloat = 0
     @State private var headerHeight: CGFloat = 0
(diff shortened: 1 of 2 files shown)


Projects/Home/Sources/Screens/ActiveHomeView.swift
```diff
@@ -98,23 +105,44 @@ struct ActiveHomeView: View {
...
             .ignoresSafeArea()
             .accessibilityHidden(true)
     }
+}
 
-    private var sheetContent: some View {
+/// The stack of cards that makes up the white sheet.
+///
+/// Its own `View` type so the seven heavy sections are behind a type boundary instead of being
+/// inlined into `ActiveHomeView.body` — see `HeroBackgroundView` for why this matters.
```
victor-reyes-by: Let's remove comments why we _extracted_ these into own views?

victor-reyes-by (approved):
Nice, I guess! 😬

PR #2618, opened 2026-09-04
feat: use server-provided actionText for termination suggestion buttons

sladan-hedvig:
## What?

Propagates the new `actionText` field on `TerminationFlowSurveyOptionSuggestion` added in [underwriter#1766](https://github.com/HedvigInsurance/underwriter/pull/1766).

- Added `actionText` to `TerminationSurveyOptionSuggestionFragment`.
- `TerminationSuggestion` gains `actionText: String?`, and `buttonTitle` returns it when present, falling back to the existing type → `L10n` mapping when it is `nil`/empty.
- Mapped the field in `TerminateContractsClientOctopus`, and gave the demo client an `actionText` so demo mode exercises the new path.
- Added `TerminationSuggestionTests` covering actionText-wins, nil-falls-back-per-type, and empty-string-falls-back.

`TerminationSurveyScreen` already renders `suggestion.buttonTitle` on the `InfoCard` button, so no view change was needed.

## Why?

Payment Protection has no tiers to move between — the thing a member can actually change is the insured amount. The button label can no longer be derived from the suggestion type alone, so the backend now sends it.

The rest of underwriter#1766 needs no iOS work: the new option ids (`BETTER_PRICE_CHANGE_AMOUNT`, `MISSING_COVERAGE_OR_TERMS_CHANGE_AMOUNT`) and the `SURVEY_OTHER_REASON` retitling flow through automatically, since the app never hardcodes survey option ids and takes all titles from the server.

## Tests

New unit tests for `TerminationSuggestion.buttonTitle`. Not verified by a build — swift-format is clean on all changed files.

## Note

The change-amount suggestions reuse the `UPGRADE_COVERAGE` / `DOWNGRADE_PRICE` types, which route into the ChangeTier flow. Worth sanity-checking that ChangeTier returns an amount option for a Payment Protection contract rather than falling into the `.emptyTier` branch.

### DIFF
--- a/Projects/TerminateContracts/Sources/Models/TerminationFlowSurveyStepModel.swift
+++ b/Projects/TerminateContracts/Sources/Models/TerminationFlowSurveyStepModel.swift
@@ -76,11 +76,13 @@ public struct TerminationRedirectionImage: Codable, Equatable, Hashable, Sendabl
 public struct TerminationSuggestion: Codable, Equatable, Hashable, Sendable {
     public let type: TerminationSuggestionType
     public let description: String
+    public let actionText: String?
     public let url: String?
 
-    public init(type: TerminationSuggestionType, description: String, url: String?) {
+    public init(type: TerminationSuggestionType, description: String, actionText: String? = nil, url: String?) {
         self.type = type
         self.description = description
+        self.actionText = actionText
         self.url = url
     }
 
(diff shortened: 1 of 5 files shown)


Projects/TerminateContracts/Sources/Models/TerminationFlowSurveyStepModel.swift
```diff
@@ -95,6 +97,9 @@ public struct TerminationSuggestion: Codable, Equatable, Hashable, Sendable {
     }
 
     public var buttonTitle: String {
+        if let actionText, !actionText.isEmpty {
```
victor-reyes-by: Nit:
Isn't `!actionText.isEmpty` too defensive?

victor-reyes-by (approved):

PR #2631, opened 2026-09-10
fix: iOS 26 SDF crash - possible corner radius issue and empty glass container

sladan-hedvig:
## Problem

Production `SIGTRAP` in `SDFStyle.distanceRange.getter` on the Home tab, iOS 26. Seen in 14.0.3 (iPhone 12, 26.3.1) and 14.0.2 (iPhone 11 Pro Max, 26.3) and on iPhone 16 Pro Max, 26.6.1 with identical stacks.

## Root cause

A corner radius **larger than the shape's geometry allows**. It only ever rendered because CoreGraphics silently clamped it. iOS 26 renders shapes through a signed-distance-field path where `SDFStyle.distanceRange` builds a `ClosedRange`; an out-of-range radius yields invalid bounds and traps instead of clamping. `cornerRadius: .infinity` is the extreme case, not a separate bug.

The limit is per edge — two adjacent radii can't exceed the edge they share. For a `RoundedRectangle` (all four corners) that means `min(w, h)`; for the `UnevenRoundedRectangle` below, where only the top corners are set and the bottom two are 0, the top radius can use the full height.

## Changes

| Site | Radius vs max valid | Fix |
|---|---|---|
| `HomeSheetContainer.surfaceShape` | 32 on a 20pt-tall shape, top corners only (max 20) | radius `20` |
| `hButton.buttonCornerModifier` | 36 on 32–56pt buttons (max 16–28) | `Capsule()` |
| `hPill` `capsuleShape` | 24 on a ~36pt pill (max ~18) | `Capsule()` |
| `CrossSellDiscountProgressComponent` ×3 | 8 in an 8pt track (max 4) | `Capsule()` |
| `hPill` wrapper | `.infinity` | `Capsule()` |

All pixel-identical to what shipped — each radius was already being clamped to the value it is now given explicitly.

Also: removed the unusable `cornerRadiusRounded` token; `isFinite` guards on `ActiveHomeView` geometry; `GlassEffectContainer` no longer built when `toolbarOptionTypes` is empty (it starts `[]`, so every cold launch wrapped a glass backdrop group around zero-sized content); removed dead `topSafeAreaInset`.

🤞

### DIFF
--- a/Projects/CrossSell/Sources/View/Components/CrossSellDiscountProgressComponent.swift
+++ b/Projects/CrossSell/Sources/View/Components/CrossSellDiscountProgressComponent.swift
@@ -19,7 +19,7 @@ struct CrossSellDiscountProgressComponent: View {
                     ForEach(1..<4) { column in
                         VStack(spacing: 4) {
                             ZStack {
-                                RoundedRectangle(cornerRadius: .cornerRadiusS)
+                                Capsule()
                                     .fill(hSurfaceColor.Opaque.secondary)
                                 if (column <= numberOfInsurances + 1) {
                                     AnimatedProgressView(orderOfExecution: column, pulse: column > numberOfInsurances)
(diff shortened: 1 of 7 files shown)


victor-reyes-by (approved):
Weird, but OK!

PR #2636, opened 2026-09-11
feat(x-sells): one cross-sell row everywhere

StylianosGakis:
WIP - https://claude.ai/code/artifact/9d05ac7b-6ebf-4be2-9369-25a9d7e2d447?sk=y8RT-Dbaa0Uwq5yDAabeew - https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1788763511277219

🤖 AI description:

The iOS half of the "X-sell lists in app 1.5" design. 126 insertions, 137 deletions.

`discountAvailable` is retired end to end. It drove three things on iOS: the button style in `CrossSellingItem`, the section header title in `CrossSellStackComponent`, and a hardcoded `true` in `CrossSellingDetent`. The field is gone from the query, the model and the client, so the storefront field can eventually be deleted rather than living on as a constant.

Per surface:

- **Home and Contracts tab** lose their cross-sell section headers. The row is always `.secondary` now.
- **Contracts tab** add-ons move from `AddonCardView` to the shared `CrossSellRow`, matching Home. The card's badge pills go with it, which matches the Figma and what Android did.
- **The sheets** gain an Addons block, and the detent leads with "For you" over "Discover more coverage". Per Figma 2.1, 2.2 and 3.1, the Addons block and the heading belong only to the plain list sheet: the modal, which leads with a recommendation, has neither.
- **Travel certificate** keeps its card, since design is still working on that screen, but its button reads "See price" so the copy is consistent everywhere.

Three things a reviewer can't get from the diff:

1. **This depends on the storefront change deploying.** Until then the Contracts tab renders backend-supplied "Save 50%" text in the new plain pill. Deploy order is backend first, then this: the backend change is safe against every shipped client and moves old builds toward the new design rather than away from it.

2. **The VoiceOver label was removed rather than updated.** `CrossSellingItem` used to pass `L10n.crossSellGetPrice` explicitly, which would now contradict a button reading "See price". `CrossSellRow` already falls back to `accessibilityAction ?? buttonTitle`, so dropping the argument makes the spoken label follow the visible one.

3. **Add-on rows are handed their tap target from outside.** Add-ons open the purchase flow rather than a store URL, and building that input needs contract data the `CrossSell` module cannot reach, so `CrossSellAddonsSection` takes an `onAddonTap` closure that `HomeTab` fills in. Neither parameter is defaulted on purpose: a sheet with rows and no handler would fail silently.

Two `Localizable.strings` lines are from `CROSS_SELL_TITLE` being ticked for ios in Lokalise; the rest of those files are untouched so the usual sync still owns them.

Verified: full `Ugglan` scheme builds clean on an arm64 simulator with all changed files force-recompiled, and `build-for-testing` on the `CrossSell` scheme compiles the four updated test files. Driven in the simulator in demo mode across Home, the Contracts tab and the sheet. Note `generic/platform=iOS Simulator` cannot build this workspace at all: the pinned `HedvigShared.xcframework` has no x86_64 slice, so a universal simulator build fails at link regardless of the code. Add-on tap navigation is not verifiable in demo mode, because the demo banner carries no contract ids; the pre-existing Home add-on row behaves identically.

### DIFF
--- a/Projects/Addons/Sources/Views/AddonCardView.swift
+++ b/Projects/Addons/Sources/Views/AddonCardView.swift
@@ -43,14 +43,14 @@ public struct AddonCardView: View {
         .hShadow(type: .custom(opacity: 0.05, radius: 5, xOffset: 0, yOffset: 4), show: true)
         .hShadow(type: .custom(opacity: 0.1, radius: 1, xOffset: 0, yOffset: 2), show: true)
         .accessibilityElement(children: .combine)
-        .accessibilityHint(L10n.voiceoverPressTo + " " + L10n.addonFlowSeePriceButton)
+        .accessibilityHint(L10n.voiceoverPressTo + " " + L10n.crossSellSeePrice)
     }
 
     var seePriceButton: some View {
         hButton(
             .small,
             .secondary,
-            content: .init(title: L10n.addonFlowSeePriceButton),
+            content: .init(title: L10n.crossSellSeePrice),
             handleSeePrice
         )
         .hButtonTakeFullWidth(true)
--- a/Projects/App/Sources/Navigation/LoggedInNavigation.swift
+++ b/Projects/App/Sources/Navigation/LoggedInNavigation.swift
@@ -769,14 +769,7 @@ struct HomeTab: View {
             presentationStyle: .detent(style: [.large]),
             options: .constant([.alwaysOpenOnTop, .withoutGrabber])
         ) { crossSells in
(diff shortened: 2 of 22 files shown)


victor-reyes-by (approved):
Nice! Just remove comments 😃

sladan-hedvig (approved):

sladan-hedvig (approved):

PR #2643, opened 2026-09-14
refactor: hold the minimum display duration through failures too

victor-reyes-by:
## [APP-XXX]

- Add `Task.withMinimumDuration(_:tolerance:_:)` to hCore
- Convert the 11 minimum-display-duration sites to it

**Behaviour change:** failures now wait out the floor too. A commit that fails on a dead network goes from ~0.2s to 3s before the error shows. Previously the hand-rolled `async let` dropped the floor whenever the call threw.

## Blockers

- Needs a product call on the failure wait

## Checklist

- [x] Unit tests pass
- [ ] Landscape verification
- [ ] iPad verification


Projects/hCore/Sources/Task.swift
```diff
@@ -5,3 +5,21 @@ extension Task {
...
+        do {
+            let result = try await block()
+            try? await Task.sleep(until: deadline, tolerance: tolerance)
+            return result
+        } catch {
+            try? await Task.sleep(until: deadline, tolerance: tolerance)
+            throw error
+        }
+    }
+}
```
victor-reyes-by, line 25: We'll wait x seconds, even on error, which imo is better, but IDK.

sladan-hedvig: I am kinda on opposite side, as soon we get error, we should throw - not wait. This is not going to happen so often anyway

sladan-hedvig (approved):

PR #2648, opened 2026-09-15
feat(SubmitClaimChat): Skriv / Spela in input for the description step

nilsbohmanstrom-hedvig:
## What

The description step (audio recording step) gets the redesigned input from Figma ([App P2 2026 › 5 · Text / voice input](https://www.figma.com/design/SogcacjzOxkCC46XcZP8lQ/App-P2-2026?node-id=2670-51082)): **Skriv · Spela in**, then a **text card** or **voice card** that takes the place of the docked input area. They replace the stacked "Spela in med röst" / "Beskriv i text" buttons, the full-screen `hTextView` editor and the voice recording detent.

Visual and interaction change only. The step still submits either free text or an uploaded recording through the same `SubmitClaimAudioStep` logic, with the same min/max length validation.

## Why

The current step asks the member to choose between two full-width buttons and then leaves the chat: text opens a full-screen editor, voice opens a sheet. Both hide the question being answered. The redesign keeps the member in the conversation: the question is scrolled into view, the input opens as a card right under it, and the answer lands in the chat as before.

## Changes

A card is a **replacement** for the docked input area, not a child of it.

- `Models/ClaimIntentStepHandler.swift`: `usesFloatingInputCard` (default `false`). While `true`, the chat fades the whole docked area – frosted panel, step input, skip button and scroll-to-bottom arrow – and shows the step's card instead.
- `Models/SubmitClaimAudioStep.swift`
  - `inputMode` (`.choose` / `.text` / `.voice`) drives `usesFloatingInputCard`; `shouldFocusTextInput` separates "tapped Skriv" from "resumed step opened pre-filled"; `submittedKind` records how the step was answered and is what the result bubble, the alignment and the payload read.
  - `beginText()` / `beginVoice()` / `cancelInput()` / `saveText()` / `saveVoice()`.
  - The length message is shown only after the member taps Skicka with too little text, and goes away once the text is valid. Skicka carries the rule as an accessibility hint.
- `Views/SubmitClaimChatScreen.swift`
  - `ClaimChatFloatingCardView` is a sibling of the docked input in the bottom `ZStack`, on any size class. It never feeds `currentStepInputHeight`, so it cannot trip `shouldHideCurrentInput`, and nothing clips its shadow.
  - `scrollToStep` scrolls the H and the question to the top when a card opens and when the text field gets focus. In the real flow the H is drawn by the task row before the question, so that row is the anchor.
  - `floatingCardView` animates on step change like the docked row does, so the card slides down and fades when the answer is sent instead of disappearing.
  - `ClaimChatDockedInputView` + `HiddenWhileFloatingCard` fade the docked area (kept in the tree so the step-to-step transitions and the measured input height stay exactly as for the other steps); `ClaimChatInputBlurBackground` is the panel extracted as-is.
  - `.claimStepErrorAlert(for:)` is the one error path, used by both the docked input and the card.
- `Views/SubmitClaimChatMessageView.swift`: `ClaimInputCardView`, the step → card registry. `ClaimStepView` is untouched.
- `Views/Steps/SubmitClaimVoiceRecordingView.swift`: the Skriv / Spela in row only. Side by side when the step is skippable, stacked otherwise.
- `Views/Steps/SubmitClaimTextCard.swift` (new): label, multiline `TextField`, length message, Avbryt / Skicka. No character counter and the message stays grey, on purpose (Figma section 2). In landscape the input and the buttons sit side by side.
- `Views/VoiceRecording/VoiceRecordingCardContent.swift`: edited in place. Form wrapper dropped, card padding/background, close button, tiles hug their content. Waveform scrubbing, VoiceOver playback action and the minutes:seconds timer layout are kept. While sending the waveform is greyed and the timer reads "Skickar…" (Figma 3.6). Recorder errors use `StateView`, centred above the controls, as the detent did.
- `Views/VoiceRecording/VoiceRecordButton.swift`: the countdown task is cancelled when the button leaves the screen.
- `Views/VoiceRecording/VoiceSendButton.swift`: disabled while sending, label reads "Skickar…".
- `Views/Components/ClaimInputCardBackground.swift` (new): `.claimInputCardBackground()` shared by both cards (`cornerRadiusXXXL`, soft shadow).
- `Models/ClaimChatScrollCoordinator.swift`: only member-driven scrolls (`isTracking` / `isDragging` / `isDecelerating`) dismiss the keyboard; programmatic scrolls and keyboard insets no longer close a card the member is typing in.
- `Extensions/View+layoutAvailability.swift`: `geometryGroupIfAvailable()` (deployment target is iOS 16).

## Test session 17 Sept

Fixed from the bug list:

- Questions land cut off when the input field is active, and the field does not scroll back to the question after scrolling up. Same cause, see `scrollToStep` above.
- Error message layout in the voice card. Back to `StateView`.
- Dismissing the voice recorder mid-countdown kept counting and vibrating. The countdown task is now cancelled with the card.
- Submitting text or audio made the card disappear without a transition. The card now slides down and fades.

## Notes on the implementation

- **Copy.** Skriv / Spela in / Skickar… use `claims_write`, `claims_record` and `CLAIMS_VOICE_RECORDING_SENDING`. The card label reuses `CLAIMS_TRIAGING_WHAT_HAPPENED_TITLE` ("Vad har hänt?").
- **Skip button.** Ghost for every step, as on master.
- **Keyboard and long questions.** The H and the start of the question are pinned at the top. The end of a very long question can sit behind the card until the member scrolls.
- **Edit.** Ändra on a text answer reopens the card prefilled (`currentFreeText`) without stealing the keyboard. Ändra on a voice answer shows the row.

## Testing

- `SubmitClaimChat` test scheme passes locally.
- Simulator, light and dark: resting row, text card with keyboard, too-short text (message appears on Skicka tap), voice card, recorder error, send and hand-off to the next step, Skriv → Avbryt → Hoppa över (skipped pill, no empty bubble), scroll back through history with the card open.
- The scroll offset from the bug list does not reproduce in the simulator. Needs a device pass.
- Not yet: iPad, small phones, VoiceOver pass.

Design: [Figma › App P2 2026 › 5 · Text / voice input](https://www.figma.com/design/SogcacjzOxkCC46XcZP8lQ/App-P2-2026?node-id=2670-51082). Section numbers above (1.1, 3.6, …) refer to the frames in that section.


Projects/SubmitClaimChat/Sources/Models/ClaimIntentStepHandler.swift
```diff
@@ -22,6 +22,8 @@ class ClaimIntentStepHandler: ObservableObject, @MainActor Identifiable {
         var animateText = true
         var isLoaderAnimating = true
         var showLoadingAnimation = true
+        /// The step's input draws its own card, so the chat hides the frosted panel behind the docked input.
+        var hidesInputPanelBackground = false
```
sladan-hedvig: Rather than a `StepUIState` flag that only hides the panel, consider putting the seam on the handler itself:

```swift
/// `true` while the member answers in a card. The chat then replaces the whole docked input
/// area - panel, skip button and scroll-to-bottom arrow included - with the card.
var usesFloatingInputCard: Bool { false }
```

One property then drives the panel, the skip button, the scroll-to-bottom arrow and the card presentation. As it stands the panel is hidden by `hidesInputPanelBackground`, the skip button by a separate `setDisableSkip(to:)` call, and the arrow isn't handled at all.

`SubmitClaimEvent.scrollToStep(id:)` below can go with it — see the `SubmitClaimAudioStep` comment.

Projects/SubmitClaimChat/Sources/Models/SubmitClaimAudioStep.swift
```diff
@@ -23,10 +23,23 @@ final class SubmitClaimAudioStep: ClaimIntentStepHandler {
         willSet {
             showTextViewOnAppear = newValue
         }
+        didSet {
+            updateSkipVisibility()
+        }
     }
     @Published var showTextViewOnAppear: Bool = false
-    @Published var isAudioInputPresented: Bool = false
+    @Published var isAudioInputPresented: Bool = false {
```
sladan-hedvig: Two independent booleans allow both cards to be open at once, and conflate "which card is showing" with "what the member answered with". Consider collapsing them to:

```swift
@Published private(set) var inputMode: InputMode = .choose   // .choose / .text / .voice
@Published private(set) var shouldFocusTextInput = false     // replaces showTextViewOnAppear
@Published private(set) var submittedKind: AudioRecordingStepType?

override var usesFloatingInputCard: Bool { inputMode != .choose }
```

`submittedKind` is the important one. `isTextInputPresented` is currently doubling as "this step was answered with text" at four call sites — `SubmitClaimVoiceRecordingResultView`, the `alignment` extension in `SubmitClaimChatMessageView`, `executeStep()` and `accessibilityValue`. That was safe while the flag was one-way, but `dismissInput()` now sets it back to false, so **Skriv → type → Avbryt → Hoppa över** renders an empty result bubble.

`shouldFocusTextInput` also separates "member tapped Skriv" (focus the field) from "resumed step opened pre-filled" (don't steal the keyboard). The current `willSet { showTextViewOnAppear = newValue }` makes those two identical.

`init` would then set `inputMode = .text` / `submittedKind = .text` for `currentFreeText` and `submittedKind = .audio` for `currentAudioUrl`, and `executeStep()` would read `let isText = inputMode == .text` instead of mutating `isAudioInputPresented`.

sladan-hedvig (commented):
The card here is a **child** of the docked input area rather than a **replacement** for it, and I think most of the friction in this PR follows from that.

Because the card is rendered inside `SubmitClaimVoiceRecordingView`, i.e. inside `CurrentStepView` inside the measured `ScrollView`, the surrounding chrome has to be neutralised piece by piece: `state.hidesInputPanelBackground` to fade the panel, `setDisableSkip(true)` to hide the skip button, `.scrollClipDisabledIfAvailable()` because the ScrollView clips the card's shadow, `.animation(value: currentStepInputHeight)` because the card changes the measured height, and a `.scrollToStep` event plus a 0.6 s timer to scroll the question up.

If instead a card stood in for the whole docked area — one overridable seam on `ClaimIntentStepHandler`, and the bottom of the screen split into two siblings (panel + step input + skip + scroll-to-bottom arrow / the floating card) — none of those five would be needed, because the card would never be inside the thing that clips, measures or docks.

That coupling also looks like a live bug — details in the comment on `SubmitClaimChatScreen.swift`.

The comments below are the concrete deltas.

Projects/SubmitClaimChat/Sources/Models/SubmitClaimAudioStep.swift
```diff
@@ -73,6 +86,45 @@ final class SubmitClaimAudioStep: ClaimIntentStepHandler {
...
+
+    /// The docked card is the whole input while it is open: Hoppa över and the frosted panel are only shown
+    /// with the Skriv / Spela in row.
+    private func updateSkipVisibility() {
+        let isCardPresented = isTextInputPresented || isAudioInputPresented
+        setDisableSkip(to: isCardPresented)
+        state.hidesInputPanelBackground = isCardPresented
+    }
+
+    private func presentInput(_ present: @escaping @MainActor (SubmitClaimAudioStep) -> Void) {
```
sladan-hedvig: `presentTextInput()` / `presentAudioInput()` / `dismissInput()` read more naturally as `beginText()` / `beginVoice()` / `cancelInput()` / `saveText()` — and `presentInputTask` plus the `ClaimChatConstants.Timing.inputCardReveal` 0.6 s sleep can go entirely.

The scroll doesn't need to be a timed side effect of opening the card; the card can own it in its own `onAppear`:

```swift
viewModel.scrollTarget = .init(id: step.id, anchor: .top)
```

That removes the `.scrollToStep` round-trip through the chat view model, and with it a bug: `ScrollTarget` is `Equatable` and the screen reacts via `onChange`, so re-opening a card on the same step writes an identical value and nothing happens. **Skriv → Avbryt → scroll up → Skriv again** doesn't scroll the question back.

Projects/SubmitClaimChat/Sources/Models/ClaimChatScrollCoordinator.swift
```diff
@@ -13,6 +13,9 @@ final class ClaimChatScrollCoordinator: ObservableObject {
     // MARK: - Published State
     @Published var isInputScrolledOffScreen = false
     @Published var shouldMergeInputWithContent = false
+    /// The inline text card keeps its keyboard open while the chat scrolls (the card scrolls the question
+    /// into view when it opens); every other step dismisses the keyboard on scroll.
+    var keepsKeyboardWhileScrolling = false
```
sladan-hedvig: `keepsKeyboardWhileScrolling` is a non-published var on the screen-wide coordinator, flipped from a card's `onAppear`/`onDisappear`. An alternative that needs no flag and no view-lifecycle dependency:

```swift
private var isScrolledByMember: Bool {
    guard let scrollView else { return false }
    return scrollView.isTracking || scrollView.isDragging || scrollView.isDecelerating
}

// Only the member scrolling dismisses the keyboard: programmatic scrolls and keyboard
// insets move the offset too, and would close a card the member is typing in.
if self?.scrollView?.viewController?.presentedViewController == nil,
   self?.isScrolledByMember == true { UIApplication.dismissKeyboard() }
```

Only a real finger-drag dismisses the keyboard. That covers every future card for free, and can't get stuck if a disappear is missed.

Projects/SubmitClaimChat/Sources/Views/SubmitClaimChatScreen.swift
```diff
@@ -129,6 +129,9 @@ public struct SubmitClaimChatScreen: View {
                             }
                     }
                     .frame(maxHeight: viewModel.currentStepInputHeight)
+                    // Grow/shrink with the input (the text card's field and validation message change height).
+                    .animation(.defaultSpring, value: viewModel.currentStepInputHeight)
```
sladan-hedvig: These two lines are symptoms of the card living inside the measured `ScrollView`, and I think there's a real bug underneath.

`currentStepInputHeight` is measured from `CurrentStepView`, which now contains the open card. `checkForScrollOffset()` computes `isInputScrolledOffScreen = neededHeight > availableHeight` with `neededHeight` = that same height. So a tall open card pushes `shouldHideCurrentInput` true — and line 115 renders the card only `if !viewModel.shouldHideCurrentInput`, so the card the member is typing in is replaced by the `ScrollToBottomButton`. The input is then short again, so it comes back. Worth trying with a long answer on a small device.

If the card were a sibling of the docked input rather than a child, it would never feed `currentStepInputHeight` and never trip `shouldHideCurrentInput` — and neither `.animation(value: currentStepInputHeight)` nor `.scrollClipDisabledIfAvailable()` would be needed.

Projects/SubmitClaimChat/Sources/Views/SubmitClaimChatScreen.swift
```diff
@@ -199,6 +200,25 @@ struct ScrollToBottomButton: View {
     }
 }
 
+/// Frosted panel behind the docked input. Hidden while the step draws its own card (see `hidesInputPanelBackground`).
+private struct CurrentStepPanelBackground: View {
```
sladan-hedvig: Instead of a panel that knows about `hidesInputPanelBackground`, consider splitting the docked area into three pieces:

- a `ClaimChatInputBlurBackground` — the frosted panel, extracted as-is
- a docked-input view — panel + ScrollView + skip + arrow, shown when `!step.usesFloatingInputCard`
- a docked-card view — the card, shown when `step.usesFloatingInputCard`, with `.padding(.horizontal/.bottom, .padding16)`, `.disabled(!step.state.isEnabled)`, `.transition(.move(edge: .bottom).combined(with: .opacity))` and the `onAppear` scroll

and adding the card view next to `currentStepView` unconditionally — a card input takes the place of the whole docked area on any size class — so it isn't gated behind `verticalSizeClass == .regular` or `shouldHideCurrentInput`.

The error handling in `CurrentStepView` is also worth extracting into a `.claimStepErrorAlert(for:)` modifier at the same time, so the card and the docked input share one explicit error path instead of the card inheriting it by being nested.

Projects/SubmitClaimChat/Sources/Views/SubmitClaimChatMessageView.swift
```diff
@@ -127,22 +127,32 @@ struct ClaimStepView: View {
...
             }
         }
         .disabled(!viewModel.state.isEnabled)
         .animation(.easeInOut(duration: 0.2), value: viewModel.state.isEnabled)
         .animation(.easeInOut(duration: 0.2), value: viewModel.state.isLoading)
         .id("step_\(viewModel.id)")
     }
+
+    // Figma "App P2 2026", section "5 · Text / voice input" 1.1: Hoppa över is a ghost button 16 pt below the Skriv / Spela in row.
+    private var isDescriptionStep: Bool { viewModel is SubmitClaimAudioStep }
```
sladan-hedvig: `viewModel is SubmitClaimAudioStep` inside the shared step view is exactly what a `usesFloatingInputCard` seam avoids. `ClaimStepView` could stay untouched — no `isDescriptionStep`, no `skipButtonStyle`, no `skipTopPadding` — with the skip button disappearing alongside the whole docked area rather than being restyled by one step type.

What would go here instead is the step → card registry:

```swift
struct ClaimInputCardView: View {
    @ObservedObject var viewModel: ClaimIntentStepHandler
    @ViewBuilder var body: some View {
        if let viewModel = viewModel as? SubmitClaimAudioStep {
            switch viewModel.inputMode {
            case .text:   SubmitClaimTextCard(viewModel: viewModel)
            case .voice:  VoiceRecordingCardContent(voiceRecorder: viewModel.voiceRecorder, onClose: ..., onSend: ...)
            case .choose: EmptyView()
            }
        }
    }
}
```

Separately: line 79 in this file still reads `isTextInputPresented` for `alignment` — should be `submittedKind == .text`.

Projects/SubmitClaimChat/Sources/Views/VoiceRecording/VoiceRecordingInlineCard.swift
```diff
@@ -0,0 +1,134 @@
+import SwiftUI
```
sladan-hedvig: Could this be an in-place edit of `VoiceRecordingCardContent.swift` instead of a new file plus a deletion? The diff would be fairly small: drop the `hForm`/`hSection` wrapper, add `.padding(.padding16)` and the card background, add an `onClose` close-button overlay, swap `StateView` for the two-line error text, and `.frame(maxWidth: 600)` → `.fixedSize(horizontal: false, vertical: true)`.

Rewriting it as a new file is what silently dropped:

- the `DragGesture` waveform scrubbing (pause-on-drag, seek-and-resume on end) — only tap-to-seek survives here
- `PlaybackAccessibilityModifier`: `.accessibilityAction { togglePlayback() }`, `.accessibilityValue(formattedTime)`, `.updatesFrequently` while recording, `.playsSound`, the playback hint. VoiceOver users currently have no way to play back their own recording.
- `DotsActivityIndicator` while sending, and the minutes:seconds timer layout with its `timerHeight` compensation

`Projects/SubmitClaimChat/CLAUDE.md:33` still references the deleted file either way.

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -15,108 +18,183 @@ struct SubmitClaimVoiceRecordingView: View {
...
+            }
+            .background(Capsule().fill(hBackgroundColor.primary))
         }
-        .transition(.opacity.combined(with: .move(edge: .bottom)))
+        .geometryGroupIfAvailable()
+        .transition(.opacity)
+    }
+
+    // MARK: - Text card
+    private var textInputCard: some View {
```
sladan-hedvig: This file is now ~300 lines holding four unrelated things. Worth moving out:

- `SubmitClaimTextInputCard` → `Views/Steps/SubmitClaimTextCard.swift`
- `SubmitClaimInputCardBackground` → `Views/Components/ClaimInputCardBackground.swift`, plus a `.claimInputCardBackground()` `View` extension so both cards read `.claimInputCardBackground()` rather than `.modifier(...)`

The voice card already lives in its own file, so the current split is inconsistent either way.

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -15,108 +18,183 @@ struct SubmitClaimVoiceRecordingView: View {
...
+        .animation(.defaultSpring, value: viewModel.isTextInputPresented)
+        .animation(.defaultSpring, value: viewModel.isAudioInputPresented)
     }
 
-    // MARK: - Initial Buttons
+    // MARK: - Initial Buttons (Skriv · Spela in)
+    // Secondary resting token, made opaque over the blurred panel (page background behind the capsule).
     private var initialButtons: some View {
-        VStack(spacing: .padding8) {
+        HStack(spacing: .padding8) {
```
sladan-hedvig: Two suggestions here:

1. Side by side only when the step is skippable, stacked otherwise — two large buttons plus a third row don't fit side by side on small devices when the skip button isn't there to absorb the height:

```swift
if viewModel.isSkippable && !viewModel.state.disableSkip {
    HStack(spacing: .padding8) { textInputButton; audioInputButton }
} else {
    VStack(spacing: .padding8) { textInputButton; audioInputButton }
}
```

2. Drop `.background(Capsule().fill(hBackgroundColor.primary))` on both buttons. Patching an opaque capsule behind a secondary button to beat the blur is really a design-system problem.

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -15,108 +18,183 @@ struct SubmitClaimVoiceRecordingView: View {
...
+}
+
+// MARK: - Text card: label · multiline field · Avbryt / Skicka (Figma section 2)
+struct SubmitClaimTextInputCard: View {
+    @ObservedObject var viewModel: SubmitClaimAudioStep
+    @EnvironmentObject var scrollCoordinator: ClaimChatScrollCoordinator
+    @FocusState private var isFocused: Bool
+    /// The min/max-length message would make the card jump on every keystroke, so it is only shown
+    /// after the user taps the disabled Skicka button, and goes away once the text is valid.
+    @State private var showsValidationError = false
```
sladan-hedvig: `showsValidationError` plus the transparent `Button` overlaid on the disabled Skicka (line 148) can both go if the error is gated at the source instead:

```swift
// Only complain once there is something to complain about - an empty card should not show an error.
textInputError = !textInput.isEmpty && characterMismatch
    ? L10n.claimsTextInputMinCharactersError(audioRecordingModel.freeTextMinLength) : nil
```

The message is then always live and an empty card simply shows nothing — no tap-to-reveal, and no `.accessibilityHidden(true)` affordance VoiceOver can't reach.

It also sidesteps a dead path in the current version: `textInputError` is only assigned in `textInput`'s `didSet`, and property observers don't fire for assignments in `init`. On a pristine card `textInput` is `""` and never written, so `characterMismatch` is true while `textInputError` is nil — tapping the disabled Skicka flips `showsValidationError` and renders nothing.

While the card is being moved to its own file, a few elements worth adding: a live `"\(count)/\(freeTextMaxLength)"` counter that turns red past the max, the error rendered in `hSignalColor.Red.element` beside it, `.accessibilityHidden(true)` on the label with an `.accessibilityHint` carrying the min-length rule, and `[weak viewModel]` in both button actions.

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -15,108 +18,183 @@ struct SubmitClaimVoiceRecordingView: View {
...
+        }
+    }
+}
+
+/// Card container shared by the text and voice inputs (Figma "Menu - iPhone": radius 32, soft shadow).
+struct SubmitClaimInputCardBackground: ViewModifier {
+    func body(content: Content) -> some View {
+        content
+            .background(
+                RoundedRectangle(cornerRadius: .cornerRadiusXXL + .padding8)
```
sladan-hedvig: `.cornerRadiusXXL + .padding8` → `.cornerRadiusXXXL`, which already exists as `32` in hCoreUI (`DefaultStyling.swift:453`).

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -188,68 +268,33 @@ struct SubmitClaimVoiceRecordingResultView: View {
...
-            isSkippable: false,
-            isRegrettable: false,
-            progress: 0.3
-        ),
-        service: .init(),
-        mainHandler: { _ in }
-    )
-    viewModel.isTextInputPresented = true
-    return VStack {
+#Preview("Text card") {
```
sladan-hedvig: Minor: this preview never sets `isTextInputPresented = true` (the "Voice card" preview does set its flag), so it renders the Skriv / Spela in row — identical to "Initial State". The prefilled `currentFreeText` isn't visible anywhere.

Projects/SubmitClaimChat/Example/Sources/SceneDelegate.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+import SubmitClaimChat
+import SwiftUI
+import hCore
+import hCoreUI
+
+/// Example app for SubmitClaimChat – runs the real claim chat against a demo client whose
```
sladan-hedvig: We can remove this file, we don't need prototype and example for this module

Projects/SubmitClaimChat/Project.swift
```diff
@@ -5,5 +5,14 @@ let project = Project.framework(
     name: "SubmitClaimChat",
     targets: Set([.framework, .example, .tests]),
     projects: ["hCore", "hCoreUI", "Claims"],
-    sdks: []
+    sdks: [],
```
sladan-hedvig: We can remove this example scripts

scripts/example-post-build-action.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
sladan-hedvig: we can also remove this file since we will no longer use it

Tuist/ProjectDescriptionHelpers/Project+Templates.swift
```diff
@@ -15,7 +15,8 @@ extension Project {
         targets: Set<FeatureTarget> = Set([.framework, .tests, .example, .testing]),
         projects: [String] = [],
         dependencies: [String] = ["CoreDependencies"],
-        sdks: [String] = []
+        sdks: [String] = [],
+        exampleScripts: [TargetScript] = []
```
sladan-hedvig: We can remove exampleScripts also

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimTextCard.swift
```diff
@@ -0,0 +1,107 @@
...
+                        .foregroundColor(hTextColor.Translucent.secondary)
+                        .transition(.opacity)
+                }
+            }
+            .animation(.defaultSpring, value: viewModel.textInputError)
+            .frame(maxWidth: .infinity, alignment: .leading)
+
+            HStack(spacing: .padding8) {
+                Spacer(minLength: 0)
+                hButton(.medium, .secondary, content: .init(title: L10n.generalCancelButton)) { [weak viewModel] in
```
sladan-hedvig: we can remove [weak viewModel]

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimTextCard.swift
```diff
@@ -0,0 +1,107 @@
...
+                    viewModel?.cancelInput()
+                }
+                .disabled(viewModel.state.isLoading)
+                hButton(.medium, .primary, content: .init(title: L10n.chatUploadPresend)) { [weak viewModel] in
```
sladan-hedvig: we can remove [weak viewModel]

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimTextCard.swift
```diff
@@ -0,0 +1,107 @@
...
+                .disabled(viewModel.characterMismatch)
+                .hButtonIsLoading(viewModel.state.isLoading)
+                .accessibilityHint(
+                    viewModel.characterMismatch
+                        ? L10n.claimsTextInputMinCharactersError(viewModel.audioRecordingModel.freeTextMinLength) : ""
+                )
+                .overlay {
+                    // A disabled button swallows taps; let a tap on the dimmed Skicka explain why it is disabled.
+                    if viewModel.characterMismatch {
+                        Button { [weak viewModel] in
```
sladan-hedvig: we can remove [weak viewModel]

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -4,120 +4,66 @@ import hCore
...
-            ) { viewModel.isTextInputPresented = true }
+    private var textInputButton: some View {
+        hButton(
+            .large,
+            .secondary,
+            content: .init(
+                title: SubmitClaimAudioStepCopy.write,
+                buttonImage: .init(image: hCoreUIAssets.penEdit.view, alignment: .leading)
+            )
+        ) { [weak viewModel] in
```
sladan-hedvig: we can remove [weak viewModel]

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -4,120 +4,66 @@ import hCore
...
-            }
+    private var audioInputButton: some View {
+        hButton(
+            .large,
+            .secondary,
+            content: .init(
+                title: SubmitClaimAudioStepCopy.record,
+                buttonImage: .init(image: hCoreUIAssets.mic.view, alignment: .leading)
+            )
+        ) { [weak viewModel] in
```
sladan-hedvig: we can remove [weak viewModel]

Projects/SubmitClaimChat/Sources/Views/SubmitClaimChatMessageView.swift
sladan-hedvig: we can remove [weak viewModel]

Projects/SubmitClaimChat/Sources/Views/SubmitClaimChatMessageView.swift
```diff
@@ -143,6 +143,33 @@ struct ClaimStepView: View {
...
+    @ViewBuilder var body: some View {
+        if let viewModel = viewModel as? SubmitClaimAudioStep {
+            switch viewModel.inputMode {
+            case .text:
+                SubmitClaimTextCard(viewModel: viewModel)
+            case .voice:
+                VoiceRecordingCardContent(
+                    voiceRecorder: viewModel.voiceRecorder,
+                    onSend: { [weak viewModel] in
+                        try await viewModel?.saveVoice()
```
sladan-hedvig: we can remove [weak viewModel]

Projects/SubmitClaimChat/Sources/Views/SubmitClaimChatMessageView.swift
```diff
@@ -143,6 +143,33 @@ struct ClaimStepView: View {
...
+                    },
+                    onClose: { [weak viewModel] in
+                        viewModel?.cancelInput()
```
sladan-hedvig: we can remove [weak viewModel]

Projects/SubmitClaimChat/Sources/Views/SubmitClaimChatScreen.swift
sladan-hedvig: we can remove [weak router]

Projects/SubmitClaimChat/Sources/Views/SubmitClaimChatScreen.swift
```diff
@@ -199,44 +261,74 @@ struct ScrollToBottomButton: View {
...
+    @EnvironmentObject var router: NavigationRouter
+
+    func body(content: Content) -> some View {
+        content
+            .onChange(of: step.state.showError) { value in
+                if value {
+                    alertVm.alertModel = .init(
+                        type: .error,
+                        message: step.state.error?.localizedDescription ?? "",
+                        action: { [weak step] in
```
sladan-hedvig, line 305: we can remove weak step

Projects/SubmitClaimChat/Sources/Views/SubmitClaimChatScreen.swift
```diff
@@ -199,44 +261,74 @@ struct ScrollToBottomButton: View {
...
+                            step?.submitResponse()
+                        },
+                        onClose: { [weak step, weak router] in
```
sladan-hedvig, line 308: we can remove weak step, weak router

Projects/SubmitClaimChat/Sources/Views/SubmitClaimChatScreen.swift
```diff
@@ -199,44 +261,74 @@ struct ScrollToBottomButton: View {
...
+                            if let claimError = step?.state.error as? ClaimIntentError {
+                                switch claimError {
+                                case .unknownStep, .unknownField:
+                                    Task { [weak router] in
```
sladan-hedvig, line 312: We can remove weak router

Projects/SubmitClaimChat/Sources/Views/VoiceRecording/VoiceRecordingCardContent.swift
```diff
@@ -2,90 +2,99 @@ import SwiftUI
 import hCore
 import hCoreUI
 
+/// Voice input for the description step (Figma "App P2 2026", section "5 · Text / voice input", section 3):
```
sladan-hedvig: WE can remove this comments

Projects/SubmitClaimChat/Sources/Views/Components/ClaimInputCardBackground.swift
```diff
@@ -0,0 +1,22 @@
+import SwiftUI
+import hCoreUI
+
+/// Card container shared by the text and voice inputs (Figma "Menu - iPhone": radius 32, soft shadow).
```
sladan-hedvig: We can remove this comment

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimTextCard.swift
```diff
@@ -0,0 +1,107 @@
+import SwiftUI
+import hCore
+import hCoreUI
+
+/// Text input for the description step (Figma "App P2 2026", section "5 · Text / voice input", section 2):
```
sladan-hedvig: We can remove this comments and generaly comments added in this PR

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -4,120 +4,66 @@ import hCore
...
-        }
-
-        var navigationTitle: String? {
-            L10n.claimsTriagingWhatHappenedTitle
-        }
-    }
+/// Copy without Lokalise keys yet – see Figma "App P2 2026", section "5 · Text / voice input", section 5 (Copy).
+enum SubmitClaimAudioStepCopy {
+    // TODO: Lokalise keys for the redesigned description step.
+    static let write = "Skriv"
```
sladan-hedvig: Replace with  L10n.claimsWrite

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -4,120 +4,66 @@ import hCore
...
+    static let record = "Spela in"
```
sladan-hedvig: Replace with L10n.claimsRecord,

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -4,120 +4,66 @@ import hCore
...
+    static let textFieldLabel = "Beskriv vad som hänt"
```
sladan-hedvig: Are we sure we want this? Can we use what we used to have: L10n.claimsTriagingWhatHappenedTitle ?

Projects/SubmitClaimChat/Sources/Views/Steps/SubmitClaimVoiceRecordingView.swift
```diff
@@ -4,120 +4,66 @@ import hCore
...
+    static let sending = "Skickar…"
```
sladan-hedvig: Added new key "CLAIMS_VOICE_RECORDING_SENDING" for this one

sladan-hedvig (commented):
_Withdrawn._

sladan-hedvig:
### Proposal: per-icon size on `hButtonImage`

The Skriv / Spela in buttons need 24pt icons (Figma "App P2 2026", section 5), but `hButton` hardcodes every icon to 16. Suggesting we add the size to the design system like this — not pushed yet, so it isn't in the diff above.

**`Projects/hCoreUI/Sources/hButton/hButton.swift`** — `hButtonImage` gains an optional `size`, defaulting to the current 16, so every existing call site is untouched:

```diff
     public struct hButtonImage: Equatable {
         let image: Image
         let alignment: HorizontalAlignment
+        let size: CGFloat
 
         public init(
             image: Image,
-            alignment: HorizontalAlignment = .center
+            alignment: HorizontalAlignment = .center,
+            size: CGFloat = .padding16
         ) {
             self.image = image
             self.alignment = alignment
+            self.size = size
         }
     }
```

and `imageView(for:)` reads it instead of the constant:

```diff
         if let image = content.buttonImage, image.alignment == alignment {
             image.image
                 .resizable()
-                .frame(width: .padding16, height: .padding16)
+                .frame(width: image.size, height: image.size)
         }
```

The icon/label gap also tightens, so the larger icon doesn't push the label too far out:

```diff
     private var mainContent: some View {
-        HStack(spacing: .padding8) {
+        HStack(spacing: .padding4) {
```

**`SubmitClaimVoiceRecordingView.swift`** — both description-step buttons then ask for 24:

```diff
-                buttonImage: .init(image: hCoreUIAssets.penEdit.view, alignment: .leading)
+                buttonImage: .init(image: hCoreUIAssets.penEdit.view, alignment: .leading, size: .padding24)
```
```diff
-                buttonImage: .init(image: hCoreUIAssets.mic.view, alignment: .leading)
+                buttonImage: .init(image: hCoreUIAssets.mic.view, alignment: .leading, size: .padding24)
```

Two things worth a reviewer's eye:

- The spacing change is app-wide, not scoped to this screen — it also affects `BankIDLoginQRView.swift:90`, `MovingFlowHouseScreen.swift:155` and `SubmitClaimDeflectView.swift:172`, the three other buttons with icons. If we'd rather not touch those, the gap can move onto `hButtonImage` as a `spacing` property defaulting to `.padding8`.
- `.padding24` fits `.large` exactly (56 − 15 − 17 = 24). `.medium` has no `minHeight` and `.small` has 32 with 14 of padding, so a caller passing a big size on those sizes would grow the button past its design-system height. An enum (`.standard` / `.large`) instead of a raw `CGFloat` would make that unrepresentable, if we want the guard rail.

nilsbohmanstrom-hedvig:
Pushed 822a195 with the review follow-ups.

On the `hButtonImage` proposal: took the enum (`.standard` 16 / `.large` 24) and the app-wide icon–label gap of 4, so the BankID login, moving-flow house and deflect buttons tighten too – on purpose.

Copy: `claims_write` / `claims_record` / `CLAIMS_VOICE_RECORDING_SENDING`, and the text card label reuses `CLAIMS_TRIAGING_WHAT_HAPPENED_TITLE`. The new key is hand-added to the .strings files for now; a `translations.sh` pull should replace it. Example app, demo root/client, `[weak]` captures and the comments are gone.

nilsbohmanstrom-hedvig:
**Scroll to question when the input card opens** (bug list: "Questions land cut off when inputfield is active" + "Input field active on scroll up")

Two bugs from the test session had the same cause. When you tap Skriv the chat scrolls the question to the top, but in the real flow every question comes after a task step and the H avatar is drawn in that row, not in the question row. So the question row landed at the top and the H was pushed off screen.

Fix: `scrollToStep` on the view model picks the row that holds the H (the task step before the question when there is one) and scrolls that to the top. The card uses it when it appears.

The text card also calls it when the field gets focus. That covers the second bug: scroll up through the history with the card open, tap the field, and the chat comes back to the question with the keyboard.

Two things you may wonder about:
- The text card fires the scroll twice on open, once on appear and once when focus lands 0.1 s later. Same target, so the second is a no-op. Voice has no focus and a resumed text answer opens without focus, so the appear scroll has to stay.
- Not verified on the simulator, the example flow can't reproduce the offset there. Needs a device pass on Monday.

nilsbohmanstrom-hedvig:
**Error message layout** (bug list: "Error message layout. Not sure if we want it like this")

The voice card showed recorder errors as two plain text lines. The detent used `StateView` with the warning icon. Back to `StateView`.

Inside the card the component's spacers stretch it to full screen, so it gets `fixedSize(horizontal: false, vertical: true)`. The title and timer moved into a `headerSection` next to `waveformSection`, and header plus waveform sit in one `ZStack` with the error, so the error is centred above the controls and the card keeps its height. Your landscape paddings are kept.

nilsbohmanstrom-hedvig:
**Dismissing the voice recorder didn't stop the countdown** (bug list: "Dismissing Voice recorder doesn't stop counter - it just keep vibrating")

The 3-2-1 countdown is a task held in `VoiceRecordButton`. Closing the card cleared `isCountingDown` on the recorder, which already stopped the recording from starting, but nothing cancelled the task, so the haptics kept going until it ran out.

Fix: the button cancels the task in `onDisappear`, same as `ClaimChatLoadingAnimationView` does with its animation task. Backgrounding the app mid-countdown keeps its current behaviour.

nilsbohmanstrom-hedvig:
**Card disappears without transition on submit** (bug list: "When submitting text or audio, it just dissapear without nice transition")

The card is only shown while there is a current step. On submit the chat clears the current step for a second while the next one loads, so the card was pulled out of the view tree instantly. Its slide-and-fade only ran for Avbryt and X.

Fix: `floatingCardView` wraps the card in the same kind of container as the docked row, animating on `currentStep?.id`. The exit is the reverse of the entrance, move from bottom plus opacity with `defaultSpring`. The inner transition on the card stays, it handles Avbryt and X where the step doesn't change.

sladan-hedvig (approved):

PR #2649, opened 2026-09-15
feat: auto submit single select steps without a default option

sladan-hedvig:
Single select steps now only show the **Confirm** button when the backend sends a `defaultSelectedId` — that's the case where there's a pre-filled answer to review. Without a default, picking an option submits the step after a short delay, and the first pick is final (taps during the pending submit are ignored).

- `requiresConfirmation` / `select(optionId:)` / `isSelectionLocked` on `SubmitClaimSingleSelectStep`; `skip()` cancels the pending submit and clears the lock.
- Confirm button wrapped in `if viewModel.requiresConfirmation`; content `VStack` now takes full width (the button used to be the only thing stretching the section).
- New `ClaimChatConstants.Timing.autoSubmitDelay` (0.5s), collapsing to 0 under `disableSubmitChatClaimAnimations`.

Not compiled locally — worth a pass on a step with and without a default, in `.pill` and `.binary` styles. The VoiceOver hint still reads "double click to select option" though it now also sends.


Projects/SubmitClaimChat/Sources/Models/SubmitClaimSingleSelectStep.swift
```diff
@@ -36,6 +54,9 @@ final class SubmitClaimSingleSelectStep: ClaimIntentStepHandler {
     }
 
     override func skip() async {
+        autoSubmitTask?.cancel()
+        autoSubmitTask = nil
+        isSelectionLocked = false
         await super.skip()
         selectedOptionId = nil
     }
```
victor-reyes-by: Should we even allow to skip, when _selection is locked_? It will never happen intentionally anyway because of short 250 ms timeout. Disable btn or just guard inside skip? 
`guard !isSelectionLocked else { return }`

Projects/SubmitClaimChat/Sources/Models/SubmitClaimSingleSelectStep.swift
```diff
@@ -4,6 +4,12 @@ import hCore
 final class SubmitClaimSingleSelectStep: ClaimIntentStepHandler {
     @Published var selectedOptionId: String?
     let model: ClaimIntentStepContentSelect
+    /// With a default selection the member confirms manually, otherwise picking an option submits the step
+    var requiresConfirmation: Bool { model.defaultSelectedId != nil }
```
victor-reyes-by, line 8: Nice, but kinda weird state when you try edit/regret your answer. In that case we have you have a previously selected option, but no "confirm" so you have to click again (if don't regret regretting 😆)

Image

victor-reyes-by (approved):
Nice. But why only for binary questions?

Projects/SubmitClaimChat/Sources/Models/SubmitClaimSingleSelectStep.swift
```diff
@@ -36,6 +54,9 @@ final class SubmitClaimSingleSelectStep: ClaimIntentStepHandler {
     }
 
     override func skip() async {
+        autoSubmitTask?.cancel()
+        autoSubmitTask = nil
+        isSelectionLocked = false
         await super.skip()
         selectedOptionId = nil
```
victor-reyes-by: Should we even allow to skip, when _selection is locked_? It will never happen intentionally anyway because of short 250 ms timeout. Disable btn or just guard inside skip? 
`guard !isSelectionLocked else { return }`

PR #2654, opened 2026-09-17
Stop the dots indicator's animation loop when its view disappears

victor-reyes-by:
## [APP-XXX]

- `.onDisappear { animate = false }` wrote a `@State` nobody reads (`animate` vs `animated`), so the dots' `asyncAfter` chain ran forever
- `.task` + throwing `Task.sleep` instead; SwiftUI cancels it on disappear

## Blockers

- None

## Checklist

- [ ] Unit tests pass
- [ ] Landscape verification
- [ ] iPad verification


sladan-hedvig (approved):
⭐  Test it on iOS 16 also 🙏

sladan-hedvig (approved):

